Issue #222, opened 2015-09-04
Unlabeled block or entity/component instantiation causes assert


mitchsm:
Omitting the required label from an entity or component instantiation (or a block) causes nvc to assert. Here's an example with an entity:

```VHDL
entity sub_ent is
end entity;

architecture a of sub_ent is
begin
end architecture a;

entity ent is
end entity;

architecture a of ent is
begin

entity work.sub_ent;

end architecture;
```

causes nvc to assert:

```
$ nvc -a unlabeled_entity.vhd 
nvc: ../src/tree.c:400: tree_ident: Assertion `item->ident != ((void *)0)' failed.

*** Caught signal 6 (SIGABRT) ***

-------- STACK TRACE --------
/lib64/libc.so.6(gsignal+0x37) [0x7f50815205d7]
/lib64/libc.so.6(gsignal+0x37) [0x7f50815205d7]
/lib64/libc.so.6(abort+0x148) [0x7f5081521cc8]
/lib64/libc.so.6(+0x2e546) [0x7f5081519546]
/lib64/libc.so.6(+0x2e5f2) [0x7f50815195f2]
nvc(tree_ident+0x72) [0x42d192]
nvc() [0x43357c]
nvc() [0x43921b]
nvc() [0x43b092]
nvc(sem_check+0x55) [0x449ac7]
nvc() [0x412f18]
nvc(main+0x463) [0x411d03]
/lib64/libc.so.6(__libc_start_main+0xf5) [0x7f508150caf5]
nvc() [0x411db9]
-----------------------------
```

I'm not sure why the backtrace is incomplete. Here's what gdb shows:

```
nvc: ../src/tree.c:400: tree_ident: Assertion `item->ident != ((void *)0)' failed.

Program received signal SIGABRT, Aborted.
0x00007ffff461b5d7 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
56	  return INLINE_SYSCALL (tgkill, 3, pid, selftid, sig);
(gdb) bt
#0  0x00007ffff461b5d7 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007ffff461ccc8 in __GI_abort () at abort.c:90
#2  0x00007ffff4614546 in __assert_fail_base (fmt=0x7ffff4764128 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
[...]

PR #230, opened 2015-09-07
Updates for Issue #222

mitchsm:
brought up-to-date


nickg:
I think this is actually a bug in the parser rather than the semantic checking phase. If you look at the grammar the labels on these concurrent statements aren't optional as they are for e.g. process statements. I borrowed your test case and reimplemented the checks in the parser in 290059f - hope that's OK!

mitchsm:
Of course it's ok! It makes more sense... glad you were able to still use the testcase.

PR #237, opened 2015-09-18
fix for issue #236

### LINKED ISSUE #236: semantic checks for missing entity actuals
The checks in sem_check_map() for missing actuals aren't consistent with the checks in sem_check_actual() for OPEN actuals. sem_check_map() requires that ports of mode other than OUT with missing actuals must have a default expression, which isn't compliant with 1.1.1.2 of the LRM. The checks in sem_check_actual() are correct.

mitchsm:
This fixes #236.

The first change at line 5501 is unrelated -- that one fixed that an attempt to map a record element caused an assertion. This change allows it to make it through to elaboration, where it gets the helpful message that that feature is not supported.


nickg:
All look OK to me, thanks.

PR #249, opened 2015-10-06
add I_DIMS to T_ENUM

mitchsm:
Hi Nick, I just wanted to get your read on this. There are 3 TODOs in bounds.c relating to enums. This change paves the way towards addressing them. Any reason not to do it this way?
thanks, Mike


nickg:
Yes this seems like a broadly good idea. There's a couple of other places that could be simplified as a result - e.g. the `T_ENUM` case in `sem_check_subtype`.

PR #254, opened 2015-10-11
Bounds check delays

mitchsm:
Add bounds-checking for rejection limits, signal assignment delays and for the delay in wait statements.


nickg:
Thanks for all your work on these. I've been really busy moving house so haven't been able to do much on nvc recently. :-(

mitchsm:
You're welcome, I'm having fun with it!

PR #256, opened 2015-10-12
Check default expressions for constants, signals, and variables

mitchsm:
Hopefully the commit log explains it all, but I've added more bounds-checking, and uncovered two bugs in the process: default values for subtypes used the base type rather than the subtype 'LEFT, and the 'TRANSACTION signal's default value was based on the original signal rather than being of type BIT.


nickg:
All looks fine except those two minor comments above.

mitchsm:
Thanks for the feedback. I've fixed both issues.

mitchsm:
Noticed that it wasn't checking port declarations, so I added that and a few tests for it.

PR #258, opened 2015-10-14
clean up interned strings

mitchsm:
Somewhat random, but I noticed there were repeats in the interned strings (elide_bounds, builtin, and last_value). Also found a few places they weren't being used.
I'm not 100% sure on the change I made to opt.c. I think it was necessary to change "last_value" to "LAST_VALUE" since it's an attribute. I think that identifier is serving two purposes now: the VHDL attribute name and the internal attribute identifier. Let me know if I got that wrong...


nickg:
I think it all looks fine apart from the last_value one which I'm also not 100% sure about. (Only have my phone at the moment so a bit hard to check.)

The purpose of the last value stuff in opt.c is to identify signals that potentially need their 'LAST_VALUE at runtime, and optimise the signals that don't.

I don't know how well this is covered by the existing unit tests. Possibly it is covered very indirectly by test_lower.c as there's a special "needs last value" operation emitted for signals tagged here.

mitchsm:
I can leave the pull-request open until you have a chance to inspect it in more detail, or I can revert that part of the change. Whatever you prefer is fine with me.

nickg:
Leave it open: I'll check it in a few days.

nickg:
I think it's fine. Sorry about the delay.

mitchsm:
No worries! I'm glad it was OK.

PR #272, opened 2015-12-04
Add more FST configure options.

hiyuh:
SSIA.


hiyuh:
then should i remove non-sensible ones completely?
or, just comment them out to keep as memorial monuments?

nickg:
I think we should just keep those two

hiyuh:
done, see 1ed29e6f3502d6574f30eabe61293618c918fa19

PR #288, opened 2016-04-23
simplify patterns

peteut:
I belive there is are two typos in `lexer.l`, `[0-9afA-f_]` should probably read `[0-9a-fA-F_]`.
Recent `flex` versions [1] also support `(?i:case_insensitive)` which allows for more compact re. 

```diff
-BIT_STRING [BOXbox]\"[0-9a-fA-f_]+\"
+BIT_STRING ?i:[box]\"[0-9a-f_]+\"
-BASED_INT  {INT}[#:][0-9a-fA-f][0-9a-fA-f_]*[#:]
+BASED_INT  {INT}[#:][?i:0-9a-f][?i:0-9a-f_]*[#:]
```
[1] flex Changelog excerpt:
```
2006-03-27  John Millaway <john43@users.sourceforge.net>

        * dfa.c, doc/flex.texi, flexdef.h, gen.c, main.c, parse.y, scan.l,
        scanflags.c, tests/test-ccl/scanner.l, tests/test-ccl/test.input: 
        Removed global variable caseins.  Added scanner stack flags for
        case-insensitivity.  Moved case-folding code from DFA-generation to
        parse time read-macros.  Added localized case-sensitivity syntax
        from Perl.  Added test for new syntax in test suite.  Documented new
        syntax.
```


nickg:
I guess the only possible problem might be on OS X where Apple ship an old version of Flex. Let's see if anyone complains...

peteut:
I just noticed that commit 7fb5768 sneaked into the pull request, which was not intended. I had to remove the m4 check as it failed for unknown reason on Travis CI. Maybe revert commit 7fb5768 for now. Sorry for that.

nickg:
Good spot! For some reason I missed that when I looked at the changes.

Issue #295, opened 2016-10-12
fetch-ieee.sh downloads HTTP URLs and does not checksum them


MikeMcQuaid:
This makes this script extremely vulnerable to MITM attacks on e.g. uncontrolled WiFi networks or the upstream servers being hacked and in either case the files being replaced with malicious versions.

MikeMcQuaid:
Suggestion: hardcode the SHA-256 checksums and check them after downloading and use an HTTPS source URL.

nickg:
This was fixed by @robertgzr above

MikeMcQuaid:
Thanks!

PR #296, opened 2016-10-12
add shasums to fetch-ieee.sh (#295)

robertgzr:


tools/fetch-ieee.sh
```diff
@@ -1,9 +1,29 @@
-#!/bin/sh
+#!/bin/bash
```
nickg: This needs to work with a normal POSIX shell for e.g. *BSD. How about we rewrite it with a new function like `download_file` that takes a URL and a SHA256. So instead of lists of files/checksums we have:

```
download_file "$ieee_url/math_complex-body.vhdl" 9dbdfaeb790f7afea854d52b84d20406717a6cff897ccc7720f10df53581c597
download_file "$ieee_url/math_complex.vhdl" 118a2c237befb730c7de09c2875a9f0cc3894e53d1d4c0c1c2779b8bc6dd0a31
download_file "$ieee_url/math_real-body.vhdl" 384516ee1ec0488a2e765c1b3cd6ab2212361f30195a46b6e044a408db4ba99a
```

tools/fetch-ieee.sh
```diff
@@ -23,30 +23,43 @@ while true; do
     esac
 done
```
nickg: Not, it's not necessary. Rest of the changes look fine: I'll merge them after you remove the prompt.

robertgzr:
Sorry. It appears I don't know how to git

PR #298, opened 2016-10-16
Abstract literals

mercstash:
Changes in lexer to handle abstract literals. Still need to handle real value specified in based literals.


mercstash:
Handles all the abstract literals as defined in LRM 13.4 of VHDL'93.

nickg:
Looks good, thanks.

PR #301, opened 2016-10-24
Adding support for quotation-mark in string literals and bit string literals

mercstash:


src/lexer.l
```diff
@@ -350,42 +352,42 @@ static int resolve_ir1045(void)
...
+  char *p = (yylval.s = xmalloc(strlen(str) + 1));
+  
+  while (*str) {
+    *p++ = toupper((int)*str);
+    str++;
+  }
+  *p = '\0';
 
-   TOKEN(tID);
+  TOKEN(tID);
```
nickg: Some of these functions only have whitespace changes - maybe your editor did that automatically? The C code in nvc should be indented with two spaces.

mercstash:
Yes, I use emacs but I don't have the right plugin for emacs, so sometimes it just do not great things... I'll review the code to reindent it manually.

nickg:
I think if you do `M-: (setq c-basic-offset 3)` in that Emacs buffer then it'll indent properly.

Also, I think if we changed the line at the top of the file to read
```
/* -*- mode: c; c-basic-offset: 3 -*- */
```
Then it should apply automatically (if your Emacs allows reading variables from files).

mercstash:
All indentation are now fixed, including deletions of tabulations.

nickg:
Great, thanks!

PR #360, opened 2017-12-26
Fixes drop_tokens_until exit condition

kraigher:
It hangs on the following input without the fix.

```vhdl

entity bug is
end entity;

architecture a of bug is
begin
  main : process is
  begin
  end;
end;
```


nickg:
Thanks for fixing this! I added a test case for it.

PR #375, opened 2018-06-01
ieee download webpage has changed

changephilip:
Hello！
Ieee download webpage has changed.
tools/fetch-ieee.sh doesn't work now.
And needed files are packed in 1076.2-1996_downloads.zip.
I tested new fetch-ieee.sh on my WSL/debian and it seemed to work well.


nickg:
Thanks for fixing this!

PR #384, opened 2019-06-14
Add support to allow JSON AST dump

svancau:


src/Makemodule.am
```diff
@@ -13,11 +13,13 @@ lib_libnvc_a_SOURCES = \
 	src/elab.c \
 	src/simp.c \
 	src/dump.c \
+	src/json.c \
 	src/eval.c \
 	src/common.c \
 	src/fbuf.c \
 	src/hash.c \
 	src/group.c \
+	src/json.c \
```
nickg: This file is listed twice

src/json.c
```diff
@@ -0,0 +1,1150 @@
...
+JsonNode *base_node = NULL;
+
+static void cannot_dump(tree_t t, const char *hint)
+{
+   printf("\n");
+   fflush(stdout);
+   fatal("cannot dump %s kind %s", hint, tree_kind_str(tree_kind(t)));
+}
+
+static void add_lineno(JsonNode *obj, tree_t node) {
```
nickg: Put the opening brace on the following line for consistency with the rest of the code

src/json.c
```diff
@@ -0,0 +1,1150 @@
...
+   json_append_member(obj, "ln", json_mknumber(loc->first_line));
+}
+
+static void add_filename(JsonNode *obj, tree_t node) {
+   const loc_t *loc = tree_loc(node);
+   json_append_member(obj, "filename", json_mkstring(istr(loc->file)));
+}
+
+/*__attribute__((format(printf,1,2)))
+static void syntax(const char *fmt, ...)
```
nickg: If this function is unused delete it rather than leaving it commented out

src/json.c
```diff
@@ -0,0 +1,1150 @@
...
+         }
+         json_append_element(range, range_obj);
+      }
+   }
+   else
+      json_append_member(type_node, "name", json_mkstring(type_name));
+   return type_node;
+}
+
+/*static void dump_op(tree_t t, int indent)
```
nickg: This is unused too

src/json.c
```diff
@@ -0,0 +1,1150 @@
...
+   json_append_member(config_node, "name", json_mkstring(istr(tree_ident(t))));
+   json_append_member(config_node, "of", json_mkstring(istr(tree_ident2(t))));
+   json_append_member(config_node, "decls", dump_decls(t));
+   return config_node;
+}
+
+void dump_json(tree_t *elements, unsigned int n_elements, const char *filename)
+{
+   FILE* dump_file = fopen(filename, "w");
+   if (!dump_file) {
```
nickg: Need to print an error here if there's a problem opening the file. Maybe just call `fatal_errno(filename)`?

src/json.c
```diff
@@ -0,0 +1,1150 @@
...
+   fflush(stdout);
+   fatal("cannot dump %s kind %s", hint, tree_kind_str(tree_kind(t)));
+}
+
+static void add_lineno(JsonNode *obj, tree_t node) {
+   const loc_t *loc = tree_loc(node);
+   json_append_member(obj, "ln", json_mknumber(loc->first_line));
+}
+
+static void add_filename(JsonNode *obj, tree_t node) {
```
nickg: I suggest you make one function `add_loc()` that stores in the filename as well as the line number to avoid any confusion

src/Makemodule.am
svancau: Fixed thanks

src/json.c
svancau: Thanks for the suggestion, I wanted to add the filename only on entities, architectures, packages and packages bodies. 
I used the assumptions that the other nodes below would only need the linenumber as the file is the same.

src/json.c
svancau: Fixed

src/json.c
svancau: Removed

src/json.c
svancau: Not sure what it was used for but I removed it. I'll put it back if required

src/json.c
svancau: Done

PR #498, opened 2022-08-01
Implicit labeling

Blebowski:
This MR adds implicit label creation. Processes and Loops are labelled according to VHDL 2008 LRM.
Statements are labelled extra.

Processes are labeled with _PX, where X starts from 0. Statements
equivalent to processes are also labeled. If equivalent process statements
exist in entity, X is not reset in architecture, but it keeps on counting.
Process counter is passed as I_IVAL attribute of T_ENTITY.

Loops are labeled with _LX, where X is reset upon each declarative
scope in which loops can occur (process body or subprogram body).

Statements are labelled with _SX, where X is reset in the same
scenarios as _LX. This-way labels of statements from different
processes do not affect each other.

Labels of statements for concurrent conditional, selected signal
assignment and concurrent assertion are preserved during simplification
pass.

The modifications are pre-condition for unique naming of statements
for statement code coverage collection.

Btw. I have received some tests failing from the CI build. Are these known
issues, or is it something I caused and I should debug them? If so, how
do I run tests locally please?

### DIFF
--- a/src/names.c
+++ b/src/names.c
@@ -1149,6 +1149,22 @@ void insert_spec(nametab_t *tab, tree_t spec, spec_kind_t kind,
    *p = s;
 }
 
+
+int* find_impl_label_loop_cnt(nametab_t *tab)
+{
+   scope_t *s = tab->top_scope;
+
+   for (; s != NULL; s = s->parent) {
+      if (s->container == NULL)
+         continue;
+
+      if (is_subprogram(s->container) || tree_kind(s->container) == T_PROCESS)
+         return &(s->lbl_cnts.loop);
+   }
+
+   return NULL;
+}
+
 ident_t get_implicit_label(tree_t t, nametab_t *tab)
 {
    int *cnt;
@@ -1164,8 +1180,7 @@ ident_t get_implicit_label(tree_t t, nametab_t *tab)
 
    case T_FOR:
    case T_WHILE:
-      // TODO: Use scop of wrapping process or subprogram
-      cnt = &(tab->top_scope->lbl_cnts.loop);
+      cnt = find_impl_label_loop_cnt(tab);
       c = 'L';
       break;
       
--- a/src/names.c
+++ b/src/names.c
@@ -110,13 +110,11 @@ typedef struct _lazy_sym {
    void       *ctx;
 } lazy_sym_t;
 
-typedef struct _label_cnts label_cnts_t;
-
-struct _label_cnts {
+typedef struct {
    int proc;
    int loop;
    int stmt;
-};
+} label_cnts_t;
 
 struct scope {
    scope_t       *parent;
@@ -1149,7 +1147,7 @@ void insert_spec(nametab_t *tab, tree_t spec, spec_kind_t kind,
    *p = s;
 }
 
-void set_impl_label_proc_cnt(tree_t t, nametab_t *tab)
+void continue_proc_labelling_from(tree_t t, nametab_t *tab)
 {
    if (t == NULL) {
       tab->top_scope->lbl_cnts.proc = 0;
@@ -1172,7 +1170,7 @@ void set_impl_label_proc_cnt(tree_t t, nametab_t *tab)
    tab->top_scope->lbl_cnts.proc = cnt;
 }
 
-int* find_impl_label_loop_cnt(nametab_t *tab)
+static inline int* find_impl_label_loop_cnt(nametab_t *tab)
 {
    scope_t *s = tab->top_scope;
 
@@ -1212,13 +1210,9 @@ ident_t get_implicit_label(tree_t t, nametab_t *tab)
       break;
    }
 
-   checked_sprintf(buf, 22, "_%C%x", c, *cnt);
+   checked_sprintf(buf, sizeof(buf), "_%C%d", c, *cnt);
    (*cnt)++;
    ident_t ident = ident_new(buf);
-
-#if 0
-   printf("Assigning implicit name: %s.%s\n", istr(scope_prefix(tab)), istr(ident));
-#endif
 
    return ident;
 }
--- a/src/names.c
+++ b/src/names.c
@@ -1149,6 +1149,28 @@ void insert_spec(nametab_t *tab, tree_t spec, spec_kind_t kind,
    *p = s;
 }
 
+void set_impl_label_proc_cnt(tree_t t, nametab_t *tab)
+{
+   if (t == NULL) {
+      tab->top_scope->lbl_cnts.proc = 0;
+      return;
+   }
+
+   assert (tree_kind(t) == T_ENTITY);
+
+   const int nstmts = tree_stmts(t);
+   int cnt = 0;
+   ident_t prefix = ident_new("_P");
+   for (int i = 0; i < nstmts; i++) {
+      tree_t stmt = tree_stmt(t, i);
+
+      // Entity statements for sure have labels since they have been implicitly
+      // labelled
+      if (ident_starts_with(tree_ident(stmt), prefix))
+         cnt++;
+   }
+   tab->top_scope->lbl_cnts.proc = cnt;
+}
 
 int* find_impl_label_loop_cnt(nametab_t *tab)
 {
@@ -1169,7 +1191,7 @@ ident_t get_implicit_label(tree_t t, nametab_t *tab)
 {
    int *cnt;
    char c;
-   char label[22];
+   char buf[22];
 
    switch (tree_kind(t)) {
    case T_PROCESS:
@@ -1190,9 +1212,15 @@ ident_t get_implicit_label(tree_t t, nametab_t *tab)
       break;
    }
 
-   checked_sprintf(label, 22, "_%C%x", c, *cnt);
+   checked_sprintf(buf, 22, "_%C%x", c, *cnt);
    (*cnt)++;
-   return ident_new(label);
+   ident_t ident = ident_new(buf);
+
+#if 0
+   printf("Assigning implicit name: %s.%s\n", istr(scope_prefix(tab)), istr(ident));
+#endif
+
+   return ident;
 }
 
 type_t resolve_type(nametab_t *tab, type_t incomplete)
--- a/src/names.c
+++ b/src/names.c
@@ -1170,21 +1170,6 @@ void continue_proc_labelling_from(tree_t t, nametab_t *tab)
    tab->top_scope->lbl_cnts.proc = cnt;
 }
 
-static inline int* find_impl_label_loop_cnt(nametab_t *tab)
-{
-   scope_t *s = tab->top_scope;
-
-   for (; s != NULL; s = s->parent) {
-      if (s->container == NULL)
-         continue;
-
-      if (is_subprogram(s->container) || tree_kind(s->container) == T_PROCESS)
-         return &(s->lbl_cnts.loop);
-   }
-
-   return NULL;
-}
-
 ident_t get_implicit_label(tree_t t, nametab_t *tab)
 {
    int *cnt;
@@ -1200,7 +1185,15 @@ ident_t get_implicit_label(tree_t t, nametab_t *tab)
 
    case T_FOR:
    case T_WHILE:
-      cnt = find_impl_label_loop_cnt(tab);
+      cnt = NULL;
+      for (scope_t *s = tab->top_scope; s != NULL; s = s->parent) {
+         if (s->container == NULL)
+            continue;
+         if (is_subprogram(s->container) || tree_kind(s->container) == T_PROCESS) {
+            cnt = &(s->lbl_cnts.loop);
+            break;
+         }
+      }
       c = 'L';
       break;
       
--- a/src/names.c
+++ b/src/names.c
@@ -110,6 +110,14 @@ typedef struct _lazy_sym {
    void       *ctx;
 } lazy_sym_t;
 
+typedef struct _label_cnts label_cnts_t;
+
+struct _label_cnts {
+   int proc;
+   int loop;
+   int stmt;
+};
+
 struct scope {
    scope_t       *parent;
    sym_chunk_t    symbols;
@@ -127,6 +135,7 @@ struct scope {
    lazy_sym_t    *lazy;
    tree_list_t    imported;
    scope_t       *chain;
+   label_cnts_t   lbl_cnts;
 };
 
 struct nametab {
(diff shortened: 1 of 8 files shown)


src/parse.c
```diff
@@ -97,6 +97,20 @@ static bool           translate_on = true;
 static nametab_t     *nametab = NULL;
 static bool           bootstrapping = false;
 
+typedef struct _label_cnts label_cnts_t;
+
+struct _label_cnts {
+   int proc;
+   int loop;
+   int stmt;
+};
```
nickg, line 106: The counters should go in the `scope_t` struct in `names.c` rather than here as that already tracks information for each scope.

src/parse.c
```diff
@@ -725,12 +739,59 @@ static ident_t loc_to_ident(const loc_t *loc)
...
+      c = 'P';
+      break;
+
+   case T_FOR:
+   case T_WHILE:
+      cnt = &(imp_label_cnts.loop);
+      c = 'L';
+      break;
+
+   case T_WAIT:
```
nickg, line 760: Make this the `default:` and get rid of `loc_to_ident` as I don't think it's used anywhere else.

src/parse.c
```diff
@@ -725,12 +739,59 @@ static ident_t loc_to_ident(const loc_t *loc)
...
+      c = 'S';
+      break;
+
+   default:
+      return loc_to_ident(loc);
+   }
+
+   char name[22];
+   memset(name, 0, sizeof(char) * 22);
+   sprintf(name, "_%C%d", c, *cnt);
```
nickg, line 784: Use `checked_sprintf` from `util.h`.

src/parse.c
```diff
@@ -725,12 +739,59 @@ static ident_t loc_to_ident(const loc_t *loc)
...
+      cnt = &(imp_label_cnts.stmt);
+      c = 'S';
+      break;
+
+   default:
+      return loc_to_ident(loc);
+   }
+
+   char name[22];
+   memset(name, 0, sizeof(char) * 22);
```
nickg, line 783: It's not necessary to zero the buffer here.

src/parse.c
```diff
@@ -7394,8 +7465,13 @@ static void p_entity_declaration(tree_t unit)
    p_entity_header(unit);
    p_entity_declarative_part(unit);
 
+   imp_label_cnts.proc = 0;
+
    if (optional(tBEGIN))
       p_entity_statement_part(unit);
+   
+   // Hold number of implicitly labeled processes in current entity
+   tree_set_ival(unit, imp_label_cnts.proc);
```
nickg, line 7474: Don't store this in the AST. It's trivial to recompute by looping over the statements in the entity and counting how many start with `_`.

src/parse.c
```diff
@@ -9398,6 +9478,7 @@ static void p_selected_waveforms(tree_t stmt, tree_t target, tree_t reject)
          tree_set_value(tree_assoc(stmt, i), a);
 
       tree_set_loc(a, CURRENT_LOC);
+      tree_set_ident(a, get_implicit_label(a, CURRENT_LOC));
```
nickg, line 9481: Shouldn't these all be `_S0` as they will be the first statement inside the process?

src/simp.c
```diff
@@ -1187,6 +1187,9 @@ static tree_t simp_cond_assign(tree_t t)
    else {
       tree_t s = tree_new(T_IF);
       tree_set_loc(s, tree_loc(t));
+      ident_t label = tree_ident(t);
+      if (label != NULL)
```
nickg, line 1191: Use `tree_has_ident` to check if the ident is null or not. On a debug build `tree_ident` will assert if it's null so you probably want to configure with `--enable-debug --disable-lto` while developing.

src/tree.c
```diff
@@ -27,7 +27,7 @@
 
 static const imask_t has_map[T_LAST_TREE_KIND] = {
    // T_ENTITY
-   (I_IDENT | I_PORTS | I_GENERICS | I_CONTEXT | I_DECLS | I_STMTS),
+   (I_IDENT | I_PORTS | I_GENERICS | I_CONTEXT | I_DECLS | I_STMTS | I_IVAL),
```
nickg, line 30: As above, don't need this.

nickg (changes requested):
> 
> Btw. I have received some tests failing from the CI build. Are these known issues, or is it something I caused and I should debug them? If so, how do I run tests locally please?

All the tests should pass on every commit, and failures need to be fixed. In this case I think it's just a few instances of `_line` in the expected output. You can run the tests with `make check` (see the "Testing" section of the README ;-).

src/parse.c
Blebowski: Done, I moved all counters to be per-scope in `scope_t`. It simplifies the parsing as it avoids state dependency
within the parser itself. OTOH, now each scope has such counters, which leads to more memory usage. Further,
some of the counters are unused, leading to waste of memory (e.g. each loop scope has loop counter, but it
is unused by numbering since loops are using counter from surrounding process or subprogram). 

It might be improved by keeping only two counters since per scope:
1. Entity, architecture scope contains only statements to be converted to processes, so counter for _P is enough.
2. Scopes within subprogram or process do not require processes to be counted, so they only need two distinct counters for _L and _S.

Note that the modified code now does not linearly label statements within loop scopes of processes, it restarts
the counter again.

src/parse.c
Blebowski: Done, code moved to `get_implicit_label` in `names.c`. There was only remaining, use of it, now commented
with TODO, is there a need to keep name of sequence of statements under case alternative?

src/parse.c
Blebowski: removed.

src/parse.c
Blebowski: Used.

src/parse.c
Blebowski: I removed it and I do the search in `set_impl_label_proc_cnt`. Maybe the function could be generic with
something like: `get_stmts_matching_prefix`, and then accessing the counter in `p_architecture_body`,
but I wanted to avoid accessing directly the scope counter from `parse.c`

src/parse.c
Blebowski: IMHO no. I tried:

`with clk_sys select clk_n <=
        '0' when '1',
        '1' when others;`

which leads to something like:

`P3: process is
    begin
      _S0: case CLK_SYS is
        when '1' =>
          _S1: CLK_N <= reject 0 ps inertial '0';
        when others =>
          _S2: CLK_N <= reject 0 ps inertial '1';
      end case;
      wait on CLK_SYS;   -- static
    end process;`

In that case, there is one _SX for case statement, and one for each signal assign for each alternative.
This is the same as when case statement is parsed.

src/simp.c
Blebowski: Done, thanks, I will try with the debug build.

src/tree.c
Blebowski: Removed now, as discussed, _PX from X within architecture is counted from referenced entity.

src/parse.c
Blebowski: Btw. how do I run only particular test?

src/names.c
```diff
@@ -110,6 +110,14 @@ typedef struct _lazy_sym {
    void       *ctx;
 } lazy_sym_t;
 
+typedef struct _label_cnts label_cnts_t;
```
nickg, line 113: You don't need to forward-declare the struct if it doesn't have any self-references. You can just declare it like

```c
typedef struct { .... } label_cnts_t;
```

src/names.c
```diff
@@ -1145,6 +1154,80 @@ void insert_spec(nametab_t *tab, tree_t spec, spec_kind_t kind,
    *p = s;
 }
 
+void set_impl_label_proc_cnt(tree_t t, nametab_t *tab)
```
nickg, line 1157: This name is quite obscure: to me "impl" means "implementation". How about `continue_labelling_from()`?

src/names.c
```diff
@@ -1145,6 +1154,80 @@ void insert_spec(nametab_t *tab, tree_t spec, spec_kind_t kind,
...
+
+      // Entity statements for sure have labels since they have been implicitly
+      // labelled
+      if (ident_starts_with(tree_ident(stmt), prefix))
+         cnt++;
+   }
+   tab->top_scope->lbl_cnts.proc = cnt;
+}
+
+int* find_impl_label_loop_cnt(nametab_t *tab)
```
nickg, line 1180: It's local to this file so it should be static, but it's only used in one place so why not inline it directly in the function below?

src/names.c
```diff
@@ -1145,6 +1154,80 @@ void insert_spec(nametab_t *tab, tree_t spec, spec_kind_t kind,
...
+      cnt = &(tab->top_scope->lbl_cnts.stmt);
+      c = 'S';
+      break;
+   }
+
+   checked_sprintf(buf, 22, "_%C%x", c, *cnt);
+   (*cnt)++;
+   ident_t ident = ident_new(buf);
+
+#if 0
```
nickg, line 1224: Don't leave debugging prints in the code

src/names.c
```diff
@@ -1145,6 +1154,80 @@ void insert_spec(nametab_t *tab, tree_t spec, spec_kind_t kind,
...
+      c = 'L';
+      break;
+      
+   default:
+      cnt = &(tab->top_scope->lbl_cnts.stmt);
+      c = 'S';
+      break;
+   }
+
+   checked_sprintf(buf, 22, "_%C%x", c, *cnt);
```
nickg, line 1220: Why `%x`? The LRM says decimal.

Use `sizeof(buf)` instead of repeating 22 here: it's easy to forget to update this if the buffer size changes later.

```c
checked_sprintf(buf, sizeof(buf), "_%c%d", c, (*cnt)++);
```

src/parse.c
```diff
@@ -7396,7 +7377,7 @@ static void p_entity_declaration(tree_t unit)
 
    if (optional(tBEGIN))
       p_entity_statement_part(unit);
-
+
```
nickg, line 7380: Trailing whitespace

src/parse.c
```diff
@@ -9012,7 +8995,8 @@ static void p_case_statement_alternative(tree_t stmt)
    consume(tASSOC);
 
    tree_t b = tree_new(T_SEQUENCE);
-   tree_set_ident(b, loc_to_ident(CURRENT_LOC));
+   // TODO: Do we really need identifier here??
+   //tree_set_ident(b, loc_to_ident(CURRENT_LOC));
```
nickg, line 8999: If the tests pass without this then it's unnecessary and you can just delete it.

src/simp.c
```diff
@@ -1187,6 +1187,9 @@ static tree_t simp_cond_assign(tree_t t)
    else {
       tree_t s = tree_new(T_IF);
       tree_set_loc(s, tree_loc(t));
+      
+      if (tree_has_ident(t))
+         tree_set_ident(s, tree_ident(t));
```
nickg, line 1192: In what circumstances is/isn't the label set here?

src/parse.c
nickg: > OTOH, now each scope has such counters, which leads to more memory usage. Further, some of the counters are unused, leading to waste of memory (e.g. each loop scope has loop counter, but it is unused by numbering since loops are using counter from surrounding process or subprogram).

Is it significant? Those counters are 12 bytes per scope. The scopes are freed when the parser exits them. What's a reasonable worst-case nesting depth - 10, 20? 240 bytes is nothing.

> 
> Note that the modified code now does not linearly label statements within loop scopes of processes, it restarts the counter again.

I don't think it matters: the LRM doesn't specify these labels (?) because you can't use them in an selected name.

> Btw. how do I run only particular test?

Easiest way is

```
./bin/run_regr <test name or sub-string>
```

Blebowski: Ok, that makes sense, thanks, I did not realize scopes are freed when parser exits them.
As for the statements, it does not matter at all as LRM says nothing there.

src/names.c
Blebowski: I removed it, srry. I am HW engineer so its been some time since I did some more intense C than for low-level CPU test in SoC RTL simulation... Therefore I loose the nuances :)

src/names.c
Blebowski: I used `continue_proc_labelling_from` to be explicit its counter of processes.

src/names.c
Blebowski: I added `static inline`, I was trying to keep the function below simple.

src/names.c
Blebowski: OK, removed. Out of curiosity, what functions should I use for debug logging?

src/names.c
Blebowski: Yep, thanks, thats a bug that slipped through.

src/parse.c
Blebowski: Removed.

src/parse.c
Blebowski: I still need to debug the tests, but the failing ones were the same with or without it.

src/simp.c
Blebowski: You are right, it is set in all circumstances, both T_COND_ASSIGN and T_COND_VAR_ASSIGN will for sure have the label set as parser underneath creates the implicit label for both of them.

src/names.c
```diff
@@ -1145,6 +1152,76 @@ void insert_spec(nametab_t *tab, tree_t spec, spec_kind_t kind,
...
+
+      // Entity statements for sure have labels since they have been implicitly
+      // labelled
+      if (ident_starts_with(tree_ident(stmt), prefix))
+         cnt++;
+   }
+   tab->top_scope->lbl_cnts.proc = cnt;
+}
+
+static inline int* find_impl_label_loop_cnt(nametab_t *tab)
```
nickg, line 1178: Ah, I actually meant just copy the body of this function to line 1208. It doesn't need to be a separate function unless the logic is required in multiple places.

nickg (commented):
I think you still need to update some of the files in `test/regress/gold`. Otherwise looks ok apart from that one comment above.

src/names.c
nickg: You can use printf or whatever, but just take them out when you commit, otherwise it looks messy. Some debug dumping is controlled by the `NVC_*_VERBOSE` environment variables.

src/parse.c
nickg: The scope structure is just used to hold some temporary data during symbol lookup and overload resolution. After parsing all the relevant data is stored in the AST (e.g. `tree_ref`)

src/names.c
nickg: I think it's easier to read as one function because you can read the whole thing linearly from top to bottom without jumping around. It's not a particularly long function and the logic is duplicated anywhere else.

Blebowski: Done, moved only to the function.

src/names.c
Blebowski: Thanks, I will keep it in mind.

src/parse.c
Blebowski: I debugged the 3 regression failures, only the expected name needed as you pointed out.

However, there is still failure in unit test `parse/guarded`. It complains about hiding `_P0` name, instead of
Q. However, IMHO `_P0` is not "hidden" since the surrounding block creates hierarchical separation and own scope. So, `_P0` will have its own hierarchical path. AFAIK you can't reffer to process in RHS the same way as to signal, so there is actually nothing to "hide".

nickg: > 
> AFAIK you can't reffer to process in RHS the same way as to signal, so there is actually nothing to "hide".

Sure there is. E.g.

```vhdl
p: x <= '1';
b: block is
begin
  p: process is
  begin
    report p'instance_name;  -- What does P refer to?
  end process;
end block;
```

The problem here is we shouldn't insert statements with implicit names into the symbol table. That didn't cause a problem before because the `_lineXX` names were unique. `p_process_statement` already handles this but where `p_concurrent_signal_assignment_statement`, `p_concurrent_assertion_statement`, etc. do

```c
insert_name(nametab, conc, NULL);
```

needs to be

```c
if (label) insert_name(nametab, conc, NULL);
```

Blebowski: OK, I see. There goes my VHDL knowledge :D The same should be the applied in `p_concurrent_procedure_call_statement`. With these, the regression passes.

src/names.c
Blebowski: Done

nickg (approved):
Looks OK now, thanks.

PR #513, opened 2022-08-12
src: Add implicit labeling of branches

Blebowski:
This PR adds implicit labeling to branches needed for implementation of
branch / decision coverage waiving.

A branch is one of:
  - T_IF conds with value - Could be covered to true / false (similarly as condition coverage now).
  - T_CASE assocs - Each assoc creates a single "branch".

Branches in Conditional signal assign, Selected signal assign, Conditional variable signal assign
are also labelled since they are converted to T_IF or T_CASE.

Branch label is generated uniquely for each branch, since it will be prefixed in the collected coverage
by name of the statement (which is unique in the given scope). Labels are numbered incrementally
when T_COND or T_ASSOC is added.

### DIFF
--- a/src/dump.c
+++ b/src/dump.c
@@ -31,6 +31,7 @@
 #define DUMP_TYPE_HINT  0
 #define DUMP_ADDRESS    0
 #define DUMP_STATICNESS 0
+#define DUMP_GEN_NAMES  0
 
 LCOV_EXCL_START
 
@@ -932,7 +933,9 @@ static void dump_stmt(tree_t t, int indent)
 
    if (tree_has_ident(t)) {
       const char *label = istr(tree_ident(t));
+#ifndef DUMP_GEN_NAMES
       if (label[0] != '_')   // Skip generated labels
+#endif
          printf("%s: ", label);
    }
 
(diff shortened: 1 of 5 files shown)


Blebowski:
Update, additional commit resets implicit statement counter with each T_CONCURRENT, since
each concurent statement will be converted to process which will contain multiple statements.
When statement counter is reset, statement names in processes from consecutive concurrent
assignments do not depend on each other.

nickg:
I'm not sure about this: VHDL doesn't have a concept of conditions having labels. That's just something we're adding as an implementation detail for coverage, so it doesn't seem right to be generating these in the parser. Can't we identify all conditions uniquely by the pair of (hierarchical path to container, tag number)? Where tag number is the index into the array of counters.

Let's defer this until you've implemented some of the other coverage changes: I think I'd need to see those first to decide if this is necessary or not. The statement numbering fix and `DUMP_GEN_NAMES` can be submitted separately now.

Blebowski:
Thanks,

I thought about this approach, and I first did it during "lower" pass when cover tag was created. When BRANCH tag was generated for each "cond" or "assoc", _BX was suffixed to the hierarchy path of the T_IF or T_CASE statement, directly into "hier" identifier of that tag. But it seemed ugly to handle part of labeling in parser, and part only during "coverage emit".

I will go on with the coverage then, and it is not an issue to rework it later.

Blebowski:
Closing this, and opening another PR with WIP of code-coverage.

PR #535, opened 2022-09-18
Code coverage

Blebowski:
Hello,

this is first PR with reworked code coverage, it is not yet to be merged, but ready for review.

Code coverage is now split into: Statement, Branch and Toggle coverage. Each can be enabled separately.
After simulation, coverage is dumped into "covdb" file.

There is a new command `-c` which allows merging multiple "covdb" files into single one, and generate
report from the merged file. Coverage report is generated "per-hierarchy" and allows navigating through
the design.

Currently, tests for coverage were removed, since the new coverage implementation significantly changes
the original implementation.

TBD before merge:
 - Review and processing of all comments
 - Fixing a bug with toggle coverage on "out" port of sub-instances -> so far crashes, I don't know why.
 - Extending HTML documentation with new command options.

TBD after merge (further improvements):
 - Support merging when the hierarchy path is not exactly matching any tag, place the tag to correct location in hierarchy.
 - Encode tags in coverage DB file "differentially" (don't store whole hierarchy path) to reduce size of covdb file.
 - Enhance Coverage report. When clicking on each coverage element, display source file where it is (in side panel) and highlight the statement/branch/signal.
 - Add "Coverage waiver file" which allows excluding statements/branches/toggles.
 - Add option to enable coverage dumping only on some part of the hierarchy (the same as wave-opt).
 - Add option to count X,Z -> 0/1  as valid transitions for toggle coverage
 - Add limit on size of signals for which toggle coverage is measured (to reduce performance impact). Typically good to avoid measuring toggle coverage on large memories.
 - Add "Expression" coverage as next coverage type. This will track combinations of input operands on logic expressions such as AND, OR, XOR, etc...
 - Add option to automatically exclude toggle coverage on "ports" when they are driven by constant value.


Blebowski:
Toggle coverage bug:

```
====================================================

** Note: initialising [10ms +23412kB]
** Note: loading top-level unit [7ms +5100kB]
** Note: elaborating design [3ms +516kB]
** Note: generating intermediate code [2ms +0kB]
** Note: dumping coverage data [1ms +320kB]
** Note: saving library [0ms +0kB]
** Note: wrote LLVM IR to /DOKUMENTY/Projekty/nvc/cover_debug/sha-512-eval/build/spect_tb/_SPECT_TB.TB_TOP.elab.1.initial.ll
** Note: wrote LLVM IR to /DOKUMENTY/Projekty/nvc/cover_debug/sha-512-eval/build/spect_tb/_SPECT_TB.TB_TOP.elab.0.initial.ll
** Note: wrote LLVM IR to /DOKUMENTY/Projekty/nvc/cover_debug/sha-512-eval/build/spect_tb/_SPECT_TB.TB_TOP.elab.0.final.ll
** Note: wrote LLVM IR to /DOKUMENTY/Projekty/nvc/cover_debug/sha-512-eval/build/spect_tb/_SPECT_TB.TB_TOP.elab.1.final.ll
** Note: code generation for 39 units using 2 threads [118ms +24936kB]
** Note: linking shared library [1ms +0kB]
nvc: ../src/rt/model.c:2965: x_map_signal: Assertion `dst_n->net == NULL' failed.

*** Caught signal 6 (SIGABRT) ***

[0x55d6dd02dad0] ../src/util.c:805 signal_handler
[0x7f715bee741f] (/usr/lib/x86_64-linux-gnu/libpthread-2.31.so) 
[0x7f715bd2400b] (/usr/lib/x86_64-linux-gnu/libc-2.31.so) ../sysdeps/unix/sysv/linux/raise.c:51 raise
[0x7f715bd03858] (/usr/lib/x86_64-linux-gnu/libc-2.31.so) /build/glibc-SzIz7B/glibc-2.31/stdlib/abort.c:79 abort
[0x7f715bd03728] (/usr/lib/x86_64-linux-gnu/libc-2.31.so) /build/glibc-SzIz7B/glibc-2.31/assert/assert.c:92 __assert_fail_base.cold
[0x7f715bd14fd5] (/usr/lib/x86_64-linux-gnu/libc-2.31.so) /build/glibc-SzIz7B/glibc-2.31/assert/assert.c:101 __assert_fail
[0x55d6dd0bc805] ../src/rt/model.c:2965 x_map_signal
[0x7f71607938f8] ../rtl/dummy_module.vhd:27 SPECT_TB.TB_TOP.DUMMY_MODULE_INST.DUMMY_MODULE_NESTED_INST [VHDL]
```
 when an instance with "out" port of std_logic type is instantiated, and toggle coverage is enabled.

nickg:
This is a big patch, it's going to take some time to go through fully. Do you mind if I create some PRs for the code-coverage branch in your fork? I think that will be the quickest way to make progress.

Blebowski:
Hi,
no problem, I don't expect to get it merged right away anyway (see the bug above which occurs when signal in port map is not directly mapped to signal in above scope), so please feel-free to go ahead and create some merge requests in the forked branch.

As for the functionality, this is something I am more-less used to in commercial simulators. The format of the output report is really "raw" as I think for digital designers / verification guys, fancy CSS classes does not really matter in the end.

Also, I thought it would be better to first merge this type of "first working version", and then go-on and start adding new features (see the list of "what to add after merge") and regression tests.

nickg:
As well as the HTML report, could you also add a coverage sub-command like `--summary` that prints a text summary? That will make it much easier to do some basic testing, similar to the existing `cover1.vhd` that got deleted. It could print the percentage branch/statement coverage for each file/instance, or even just the percentage for the whole design like the old one did.

src/rt/cover.c
```diff
@@ -368,323 +527,574 @@ static cover_file_t *cover_file(const loc_t *loc)
...
+   if (exists) {
+      if (!S_ISDIR(path_stat.st_mode))
+         fatal_errno("Path: '%s' set as destination for coverage reports is not a folder!", path);
+   } else {
+      mkdir (path, 'w');
+   }
 
-   char *dir LOCAL = xasprintf("%s.cover", istr(name));
+   char *subdir = xasprintf("%s/hier", path);
+   mkdir(subdir, 'w');
```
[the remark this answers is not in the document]
Blebowski: Done, pushed.

Blebowski:
> As well as the HTML report, could you also add a coverage sub-command like `--summary` that prints a text summary? That will make it much easier to do some basic testing by diffing the output, similar to the existing `cover1.vhd` that got deleted. It could print the percentage branch/statement coverage for each file/instance, or even just the percentage for the whole design like the old one did.

Actually, if you run the `-c` with `-V` (the same kind of verbose as elab), you will get more verbose info. E.g. in my debug design:

```
** Note: initialising [13ms +22340kB]
** Note: Loading input coverage database: cov_dbs/tb_top_G0.covdb [0ms +0kB]
** Note: Loading input coverage database: cov_dbs/tb_top_G1.covdb [0ms +1156kB]
** Note: Loading input coverage database: cov_dbs/tb_top_G2.covdb [1ms +116kB]
** Note: Saving merged coverage database to: cov_dbs/tb_top_merged.covdb [2ms +8kB]
** Note: Generating coverage report to folder: coverage_output. [0ms +0kB]
** Note: Code coverage results on: SPECT_TB.TB_TOP
** Note:      Statement:  87.5 %
** Note:      Branch:     68.8 %
** Note:      Toggle:     8.8 %
```
I have used the `progress` as you are using for elab. Right now the coverage merging time is negligible, but with more complex designs, it might take non-trivial time, and I would like to have some option to "benchmark" it.

src/lower.c
```diff
@@ -111,7 +111,12 @@ typedef struct {
    bool        is_const;
 } map_signal_param_t;
 
-typedef void (*lower_field_fn_t)(type_t, vcode_reg_t, vcode_reg_t, void *);
+typedef struct {
+   ident_t     hier;
+   unsigned    flags;
+} cov_toggle_ctx_t;
```
nickg, line 117: You don't need to pass the flags through, you can just check if `tree_kind(where) == T_SIGNAL_DECL` or `T_PORT_DECL`.

Blebowski: Good, point, I wasn't sure whether the nesting through elements of arrays will not change the tree kind.

Blebowski:
OK, @nickg , now the thing is ready for review. I managed to elaborate and run bit of my fairly complex
design (CTU CAN FD IP core - so far I tried only RTL top, I don't have the whole TB ported to NVC) with all coverage types enabled. Btw, the performance impact is immediately seen:

Without coverage:
```
** Note: initialising [8ms +23460kB]
** Note: loading top-level unit [12ms +7264kB]
** Note: elaborating design [450ms +35640kB]
** Note: generating intermediate code [320ms +29352kB]
** Note: saving library [22ms +264kB]
** Note: code generation for 5144 units using 7 threads [25096ms +76316kB]
** Note: linking shared library [7ms +0kB]
```
But thats something what is expected.

With all coverage types enabled:
```
Elaborating:
** Note: initialising [9ms +23404kB]
** Note: loading top-level unit [11ms +7328kB]
** Note: elaborating design [454ms +35644kB]
** Note: generating intermediate code [343ms +36800kB]
** Note: dumping coverage data [20ms +264kB]
** Note: saving library [28ms +0kB]
** Note: code generation for 5144 units using 7 threads [32820ms +84024kB]
```

But

nickg:
> ** Note: code generation for 5144 units using 7 threads [32820ms +84024kB]

OK, it doesn't seem too bad. The slow code generation with LLVM is something I'm working on separately anyway. Do you know which of statement/branch/toggle coverage has the largest effect?

Blebowski:
I also expected worse, but its only elaboration, I would be curious about the simulation.

Individual coverage types have following results:

No coverage:

```
Elaborating:
** Note: initialising [16ms +23244kB]
** Note: loading top-level unit [15ms +7204kB]
** Note: elaborating design [446ms +35880kB]
** Note: generating intermediate code [236ms +29228kB]
** Note: saving library [26ms +0kB]
** Note: code generation for 5144 units [24592ms +82592kB]
** Note: linking shared library [18ms +0kB]
```

Statement only:

```
Elaborating:
** Note: initialising [14ms +23632kB]
** Note: loading top-level unit [16ms +7140kB]
** Note: elaborating design [468ms +35868kB]
** Note: generating intermediate code [256ms +31336kB]
** Note: dumping coverage data [12ms +264kB]
** Note: saving library [26ms +0kB]
** Note: code generation for 5144 units [29448ms +83380kB]
** Note: linking shared library [22ms +0kB]
```

Branch only:

```
Elaborating:
** Note: initialising [15ms +23284kB]
** Note: loading top-level unit [13ms +7268kB]
** Note: elaborating design [429ms +35836kB]
** Note: generating intermediate code [233ms +31840kB]
** Note: dumping coverage data [9ms +264kB]
** Note: saving library [27ms +0kB]
** Note: code generation for 5144 units [27165ms +81568kB]
** Note: linking shared library [21ms +488kB]
```

Toggle only:

```
** Note: initialising [8ms +23408kB]
** Note: loading top-level unit [10ms +7332kB]
** Note: elaborating design [450ms +35700kB]
** Note: generating intermediate code [247ms +34312kB]
** Note: dumping coverage data [15ms +264kB]
** Note: saving library [31ms +0kB]
** Note: code generation for 5144 units [25430ms +85972kB]
** Note: linking shared library [17ms +1052kB]
```

All coverage types:

```
Elaborating:
** Note: initialising [10ms +23404kB]
** Note: loading top-level unit [20ms +7332kB]
** Note: elaborating design [453ms +35720kB]
** Note: generating intermediate code [259ms +36628kB]
** Note: dumping coverage data [25ms +264kB]
** Note: saving library [23ms +0kB]
** Note: code generation for 5144 units [31899ms +89000kB]
** Note: linking shared library [18ms +240kB]
```

I re-ran also without coverage because last time I had docker running on side with some GHDL simulation,
so its not 100% comparable. Now all measurements have the same "background" noise.

Blebowski:
Hello @nickg do you have any news with regards to coverage? Did you happend to have a time to review it ? I am wondering whether I should start working on some of the "TODOs" from the first comment in this MR, or whether I should wait and expect some major changes due to review.

nickg:
Can you have a look at the open PR for your branch? https://github.com/Blebowski/nvc/pull/3

Blebowski:
Oh, thanks, I completely missed this-one. I probably should set some email notifications on my Github repos...

nickg:
Oh, OK, I thought it sent emails by default.

src/rt/cover.c
```diff
@@ -148,162 +120,328 @@ static bool cover_is_stmt(tree_t t)
...
+         break;
+
+      // TODO: Could merging be done more efficiently?
+      bool found = false;
+      for (int i = 0; i < tagging->tags.count; i++) {
+         cover_tag_t *old = AREF(tagging->tags, i);
+
+         // Compare based on hierarchical path, each
+         // statement / branch / signal has unique hierarchical name
+         if (!ident_compare(new.hier, old->hier)) {
```
nickg, line 341: Just compare the pointers directly

```c
if (new.hier == old.hier) { ...
```

src/rt/cover.c
```diff
@@ -148,162 +120,328 @@ static bool cover_is_stmt(tree_t t)
...
+#endif
+
+   uint32_t sig_size = signal_size(s);
+   int32_t *toggle_mask = ((int32_t *)user) + sig_size - 1;
+
+   for (int i = 0; i < sig_size; i++) {
+      uint8_t new = ((uint8_t*)signal_value(s))[i];
+      uint8_t old = ((uint8_t*)signal_last_value(s))[i];
+
+      // std_ulogic
```
nickg, line 403: Try defining an enum here:

```c
enum std_ulogic { _U, _X, _0, _1, .... };
```

src/rt/cover.c
```diff
@@ -368,323 +504,571 @@ static cover_file_t *cover_file(const loc_t *loc)
...
-      if (mask == 1)
-         hl->help = "Condition never evaluated to TRUE";
-      else if (mask == 2)
-         hl->help = "Condition never evaluated to FALSE";
+static void cover_print_hierarchy_summary(FILE *f, cover_stats_t *stats, ident_t hier,
+                                          bool top)
+{
+   char dir[6] = {0};
+   if (top)
+      sprintf(dir, "hier/");
```
nickg, line 699: `sprintf` is one of those functions like `strcpy` that should almost never be used (lots of tools like Coverity will flag it).

src/rt/cover.c
```diff
@@ -368,323 +504,571 @@ static cover_file_t *cover_file(const loc_t *loc)
...
+
+   fprintf(f, "   <tr>\n"
+              "      <td><a href=\"%s%s.html\">%s</a></td>\n",
+              dir, istr(hier), istr(hier));
```
nickg, line 703: Get rid of `dir` and just use `top ? "hier/" : ""`.

src/rt/cover.c
```diff
@@ -368,323 +504,571 @@ static cover_file_t *cover_file(const loc_t *loc)
...
+   char *hier LOCAL = xasprintf("%s/%s.html", dir, istr(ctx->start_tag->hier));
+   cover_tag_t *tag = ctx->start_tag;
 
-   cover_html_header(fp, "Coverage report for %s", f->name);
+   // TODO: Handle escaped identifiers in hierarchy path!
+   FILE *f = fopen(hier, "w");
+   if (f == NULL)
+      fatal("Failed to open file: %s\n", hier);
+
+   ctx->ch_stmt.hits = xcalloc(1024 * sizeof(cover_pair_t));
```
nickg, line 906: Use `xcalloc_array` instead of multiplying and calling `calloc` as it handles overflows (I know this won't overflow but it's a good general practice).

src/rt/cover.c
```diff
@@ -368,323 +504,571 @@ static cover_file_t *cover_file(const loc_t *loc)
...
+   return tag;
+}
 
-   char *dir LOCAL = xasprintf("%s.cover", istr(name));
 
-   lib_t work = lib_work();
-   lib_mkdir(work, dir);
+void cover_report(const char *path, cover_tagging_t *tagging)
+{
+   char *subdir = xasprintf("%s/hier", path);
```
nickg, line 1053: `subdir` leaks I think. Put `LOCAL` after the variable name and it will be freed when it goes out of scope.

src/rt/model.c
```diff
@@ -630,6 +630,21 @@ const void *signal_value(rt_signal_t *s)
    return s->shared.data;
 }
 
+uint32_t signal_size(rt_signal_t *s)
```
nickg, line 633: No need to add `signal_size` and `signal_name` - just include `rt/structs.h` and access the fields directly.

src/rt/model.c
```diff
@@ -1236,7 +1251,6 @@ static void clone_source(rt_model_t *m, rt_nexus_t *nexus, rt_source_t *old,
    case SOURCE_PORT:
       {
          new->u.port.input = old->u.port.input;
-
```
nickg, line 1239: Restore the blank line :-)

src/rt/model.c
```diff
@@ -1754,28 +1768,41 @@ static void reset_coverage(rt_model_t *m)
...
+   cover_count_tags(m->cover, &n_stmts, &n_branches, &n_toggles);
 
    int32_t *cover_stmts = jit_find_symbol(m->jit, ident_new("cover_stmts"));
    if (cover_stmts != NULL)
       memset(cover_stmts, '\0', sizeof(int32_t) * n_stmts);
 
-   int32_t *cover_conds = jit_find_symbol(m->jit, ident_new("cover_conds"));
-   if (cover_conds != NULL)
-      memset(cover_conds, '\0', sizeof(int32_t) * n_conds);
+   int32_t *cover_branches = jit_find_symbol(m->jit, ident_new("cover_branches"));
```
nickg, line 1784: Rebase and replace this with `ffi_find_symbol`.

src/rt/model.c
```diff
@@ -1853,7 +1880,6 @@ void model_reset(rt_model_t *m)
    MODEL_ENTRY(m);
 
    // Initialisation is described in LRM 93 section 12.6.4
-
```
nickg, line 1856: And this one ;-).

src/rt/model.c
```diff
@@ -3274,3 +3300,10 @@ void x_unreachable(tree_t where)
    else
       jit_msg(NULL, DIAG_FATAL, "executed unreachable instruction");
 }
+
+void x_cover_setup_toggle_cb(sig_shared_t *ss, int32_t *toggle_mask)
```
nickg, line 3304: This can go in `cover.c`, no need to add it here.

src/rt/model.h
```diff
@@ -25,6 +25,7 @@ rt_model_t *model_new(tree_t top, jit_t *jit);
 void model_free(rt_model_t *m);
 void model_reset(rt_model_t *m);
 void model_run(rt_model_t *m, uint64_t stop_time);
+void model_finish(rt_model_t *m);
```
nickg, line 28: Not used any more?

nickg (changes requested):

Blebowski:
Hi @nickg , thanks for review, I will look at all the stuff (including your last MR), but I am swamped by my regular work duties these days, so its going to take some time.

src/lower.c
```diff
@@ -5114,7 +5231,6 @@ static void lower_signal_assign_target(target_part_t **ptr, tree_t where,
          vcode_reg_t data_reg = lower_array_data(lower_resolved(type, src_reg));
          vcode_reg_t count_reg = lower_scalar_sub_elements(type, p->reg);
          vcode_reg_t nets_raw = lower_array_data(p->reg);
-
```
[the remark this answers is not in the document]
Blebowski: Ok, went through the diff, I hope I caught all of them.

src/nvc.c
```diff
@@ -251,6 +251,26 @@ static int elaborate(int argc, char **argv)
...
+               else if (*optarg == 't')
+                  opt_set_int(OPT_COVER_TOGGLE, 1);
+               else if (*optarg == 'b')
+                  opt_set_int(OPT_COVER_BRANCH, 1);
+               else if (*optarg != ',')
+                  fatal("Unknown coverage type '%s'. Valid coverage types are: \n"
+                        "  s (statement)\n"
+                        "  t (toggle)\n"
+                        "  b (branch)\n"
+                        "Selected coverage types shall be comma separated. E.g.:\n"
```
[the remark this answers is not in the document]
Blebowski: True, checking should be better, but I would keep it comma separated to keep the door open for commands which are not single letter.

src/rt/cover.c
```diff
@@ -19,48 +19,68 @@
...
 
-#define SUB_COND_BITS 5
-#define MAX_SUB_CONDS (1 << SUB_COND_BITS)
-#define SUB_COND_MASK (MAX_SUB_CONDS - 1)
+//#define COVER_DEBUG
 
-#define PERCENT_RED    50.0f
-#define PERCENT_ORANGE 90.0f
+#define MARGIN_LEFT "400"
+#define SIDEBAR_WIDTH "370"
```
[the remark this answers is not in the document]
Blebowski: OK, I have put it to 30% fixed always. I have tried to experiment with CSS `clamp` function, but
it gave me weird behavior. I don't think it matters that much as long as the UI is usable. Also,
I don't really like front-end web stuff, so I try to avoid it...

src/nvc.c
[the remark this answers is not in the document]
Blebowski: I implemented something more complex, but maybe its too dirty. Some "generic" wrapper for command line processing above "opt" could be handy.

src/nvc.c
```diff
@@ -858,6 +883,72 @@ static int dump_cmd(int argc, char **argv)
...
+      fatal("No input coverage database FILE specified");
+
+   cover_tagging_t *cover = NULL;
+
+   // Rest of inputs are coverage input files
+   for (int i = optind; i < argc; i++) {
+      progress("Loading input coverage database: %s", argv[i]);
+      fbuf_t *f = fbuf_open(argv[i], FBUF_IN, FBUF_CS_NONE);
+
+      if (i == optind) {
```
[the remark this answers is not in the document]
Blebowski: Went through the code and fixed everything I could find.

src/nvc.c
```diff
@@ -941,8 +1044,8 @@ static void usage(void)
           "Run options:\n"
           "     --dump-arrays\tInclude nested arrays in waveform dump\n"
           "     --exclude=GLOB\tExclude signals matching GLOB from wave dump\n"
-          "     --exit-severity=\tExit after assertion failure of "
-          "this severity\n"
+          "     --exit-severity=\tExit after assertion failure of \n"
+          "                     \tthis severity\n"
```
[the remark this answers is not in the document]
Blebowski: I changed it back. The reason for change was confusing look in the terminal once `--help` was used.

src/rt/cover.c
```diff
@@ -148,162 +120,328 @@ static bool cover_is_stmt(tree_t t)
...
+cover_tag_t *cover_add_tag(tree_t t, ident_t hier, cover_tagging_t *ctx,
+                           tag_kind_t kind, uint32_t flags)
 {
-   cover_tagging_t *ctx = xcalloc(sizeof(cover_tagging_t));
-   ctx->tree_hash = hash_new(1024);
+   int *cnt;
+
+   assert (ctx != NULL);
+
+   if (kind == TAG_STMT) {
```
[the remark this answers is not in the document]
Blebowski: Done

src/rt/cover.c
Blebowski: True, good point.

src/rt/cover.c
Blebowski: Done.

src/rt/cover.c
Blebowski: OK, I see the point, will stick to `checked` version of it, despite not being able to cause seg-fault in this concrete case. I don't do production grade SW, so I lack this type of knowledge. In RTL though, there are several
(conceptually similar) topics I insist on.

src/rt/cover.c
Blebowski: Done

src/rt/cover.c
Blebowski: Done

src/rt/cover.c
Blebowski: True, added the flag. Btw. does that work accross compilers or is it GCC specific (Google says its GCC only). That brings me to a question: Why are you using GCC to build NVC while you use LLVM as code generation engine? Wouldn't it be better to use just single compiler?

src/rt/model.c
Blebowski: Done

src/rt/model.c
Blebowski: Done

src/rt/model.c
Blebowski: Done

src/rt/model.c
Blebowski: OK, moved

src/rt/model.h
Blebowski: Removed

src/rt/model.c
Blebowski: Done

src/rt/model.h
Blebowski: Removed

src/rt/model.c
Blebowski: OK, moved

Blebowski:
I have processed most of the remarks, still need to fix tests.

I completely removed implicit labeling of branches and statements from parser, and moved it all
to the coverage implementation to have it better decoupled from rest of the code (as you suggested).
Looking at it, now it makes more sense.

Further, I nest coverage scope on almost every tree item, it is easier to handle in the code, and it makes
the statements within large processes independed of each other (since they are not linearly labelled).

Blebowski:
Removal of implicit statement labelling completely from parser, had only one issue in `p_component_instantiation` statement. I worked around by not inserting NULL into symbol table. Since parse error is thrown in this case,
IMHO its unimportant anyway, but I am curious about your opinion.

The same applies for procedure call in `test_parse`, since there is no implicit labelling, assertions which
relied on either `_SX` or old `__lineXYZ` label existing were failing. I removed them since the labels do
not exist anymore.

Let me know whether it is OK, like this, or you would like to return the previous "__lineXYZ" labelling.
I would try to avoid keeping some part of the `_SX` labelling in parser, when the same thing is done
in `cover.c`.

nickg:
> 
> Let me know whether it is OK, like this, or you would like to return the previous "__lineXYZ" labelling. I would try to avoid keeping some part of the `_SX` labelling in parser, when the same thing is done in `cover.c`.

I think this is OK but could you make a separate PR for it?

Blebowski:
OK, I will do that. Btw. WBU the failing Ubuntu test? Should I implement JIT-IIR support for cover VCODE ops?
Also, what is the architecture behind the JIT? I understand that you want to speed-up the code generation, but
I am somehow lost on the flow of data / compilation (I have practically zero experience with JIT compilers).
Is the main idea to replace the LLVM API with directly emmited LLVM IR (or is it custom IR) to make it faster?
Any "lecture for dummies" is appreciated :)

src/nvc.c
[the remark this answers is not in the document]
nickg: This is fine but see Blebowski/nvc#5 and also please split this into a separate function

```c
static cover_mask_t parse_cover_mask(const char *str)
```

src/nvc.c
```diff
@@ -252,6 +252,41 @@ static int elaborate(int argc, char **argv)
          break;
       case 'c':
          opt_set_int(OPT_COVER, 1);
+         if (optarg) {
+            char prev = 0;
+            int n_chars = 0;
+            const char *full_cov_opt = optarg;
+            do {
+               printf("Processing coverage type %c\n", *optarg);
```
nickg, line 260: Remove this printf

src/nvc.c
```diff
@@ -252,6 +252,41 @@ static int elaborate(int argc, char **argv)
...
+               if (*optarg == ',' || *optarg == '\0') {
+                  if (prev == 's')
+                     opt_set_int(OPT_COVER_STMT, 1);
+                  else if (prev == 't')
+                     opt_set_int(OPT_COVER_TOGGLE, 1);
+                  else if (prev == 'b')
+                     opt_set_int(OPT_COVER_BRANCH, 1);
+                  else
+                     fatal("Unknown coverage type '%c'. Valid coverage types are: \n"
```
nickg, line 269: Error messages ought to start with a lower case letter to match the style of the rest of the program. If you want to provide additional information you can use the `diag_*` functions like:

```c
diag_t *d = diag_new(DIAG_FATAL, NULL);
diag_printf(d, "unknown coverage type '%c'", *optarg);
diag_hint(d, NULL, "valid coverage types are: \n"
          "  s (statement)\n"
          "  t (toggle)\n"
          "  b (branch)");
diag_hint(d, NULL, "selected coverage types shall be "
          "comma separated e.g $bold$--cover=s,t,b$$");
diag_emit(d);
fatal_exit(EXIT_FAILURE);
```

nickg:
> WBU the failing Ubuntu test? Should I implement JIT-IIR support for cover VCODE ops?

Don't worry about that for now, I'll fix it up later. If you restore the `cover` flag for `cover1` in `testlist.txt` it should get skipped in that mode.

> Also, what is the architecture behind the JIT? I understand that you want to speed-up the code generation, but I am somehow lost on the flow of data / compilation (I have practically zero experience with JIT compilers). Is the main idea to replace the LLVM API with directly emmited LLVM IR (or is it custom IR) to make it faster? Any "lecture for dummies" is appreciated :)

Previously the flow looked like:

```
VHDL --> vcode --> LLVM ahead-of-time for runtime
              \--> Evaluator for constant folding / elaboration
```

And in the future it's going to be

```
VHDL ----\                      /---> LLVM ahead-of-time for IEEE, etc. libraries
Verilog ---> vcode ---> JIT IR -----> LLVM JIT for frequently executed code
PSL -----/                      |---> Interpreter for elaboration / infrequently executed
...                             \---> (Maybe) lightweight code generator like GHDL mcode
```

The problems with the old approach are:
* The evaluator and the LLVM vcode backends were totally separate so every new vcode operation had to be implemented in two places. And historically the evaluator was both slow and buggy (sometimes gave different results if something was evaluated at elaboration time vs runtime). I also want to do some refactoring of how vcode works and this doubles the number of consumers that need updating.
* All the generated code has to be passed through LLVM which is very slow (because it's designed to generate high quality code). But a lot of VHDL code is either very simple so it doesn't benefit from the analysis LLVM does or is only executed once (e.g. initialisation functions).
* I get the impression that most users re-elaborate the design every time they run it (e.g. `nvc -e foo -r`). That's different to normal software compilation where the program is compiled once and run thousands of times.

The new JIT IR is designed to be quite low level (it looks a bit like a simplified RISC assembly language) so it shouldn't have to change often and is also easy to write an efficient interpreter for. So now there's one pass that converts vcode IR into JIT IR (`src/jit/jit-irgen.c`) and that can either be interpreted (`jit-interp.c`) or compiled via LLVM (`jit-llvm.c`). The latter isn't used at the moment unless you pass a special configure flag, but eventually it will replace the existing vcode->LLVM translation in `cgen.c`.

Blebowski:
> > WBU the failing Ubuntu test? Should I implement JIT-IIR support for cover VCODE ops?
> 
> Don't worry about that for now, I'll fix it up later. If you restore the `cover` flag for `cover1` in `testlist.txt` it should get skipped in that mode.
OK

> 
> > Also, what is the architecture behind the JIT? I understand that you want to speed-up the code generation, but I am somehow lost on the flow of data / compilation (I have practically zero experience with JIT compilers). Is the main idea to replace the LLVM API with directly emmited LLVM IR (or is it custom IR) to make it faster? Any "lecture for dummies" is appreciated :)
> 
> Previously the flow looked like:
> 
> ```
> VHDL --> vcode --> LLVM ahead-of-time for runtime
>               \--> Evaluator for constant folding / elaboration
> ```
> 
> And in the future it's going to be
> 
> ```
> VHDL ----\                      /---> LLVM ahead-of-time for IEEE, etc. libraries
> Verilog ---> vcode ---> JIT IR -----> LLVM JIT for frequently executed code
> PSL -----/                      |---> Interpreter for elaboration / infrequently executed
> ...                             \---> (Maybe) lightweight code generator like GHDL mcode
> ```
> 
> The problems with the old approach are:
> 
> * The evaluator and the LLVM vcode backends were totally separate so every new vcode operation had to be implemented in two places. And historically the evaluator was both slow and buggy (sometimes gave different results if something was evaluated at elaboration time vs runtime). I also want to do some refactoring of how vcode works and this doubles the number of consumers that need updating.
> * All the generated code has to be passed through LLVM which is very slow (because it's designed to generate high quality code). But a lot of VHDL code is either very simple so it doesn't benefit from the analysis LLVM does or is only executed once (e.g. initialisation functions).

I see, that makes sense, doing "on the flight" generation on infrequently invoked parts of the code is IMHO good
option. Also, it allows recompiling the code if some of it needs to change, e.g. due to increased visibility of some of its parts for debug. AFAIK, most commercial tools do it the same way. Once you launch the tool (post elab), you can "enable" dumping part of the design which on-the-fly recompile that part of the design and insert hooks into the code
upon signal change.



> * I get the impression that most users re-elaborate the design every time they run it (e.g. `nvc -e foo -r`). That's different to normal software compilation where the program is compiled once and run thousands of times.
 
That's true since most of the time the code is launched, its being debugged, so you need to change it and re-analyze.
For regression testing, you can avoid doing elaboration multiple times in some cases (e.g. if its possible to pass a seed to randomization initialization at run-time, and you run the same test with multiple randomization seeds). However,
if you pass e.g. test name via top level generic, you need to re-elaborate the whole TB each time.

> The new JIT IR is designed to be quite low level (it looks a bit like a simplified RISC assembly language) so it shouldn't have to change often and is also easy to write an efficient interpreter for. So now there's one pass that converts vcode IR into JIT IR (`src/jit/jit-irgen.c`) and that can either be interpreted (`jit-interp.c`) or compiled via LLVM (`jit-llvm.c`). The latter isn't used at the moment unless you pass a special configure flag, but eventually it will replace the existing vcode->LLVM translation in `cgen.c`.

Thanks for explanation. Btw. aren't you afraid that due to inserting one more IIR into the process, the elaboration is going to be even slower in the end?

src/nvc.c
Blebowski: Changed to the code you proposed.

src/nvc.c
Blebowski: srry, forgot about that one.

Blebowski:
On my side the MR is ready, is there something else I should look at on it?

src/nvc.c
```diff
@@ -251,7 +252,42 @@ static int elaborate(int argc, char **argv)
          opt_set_str(OPT_DUMP_VCODE, optarg ?: "");
          break;
       case 'c':
-         opt_set_int(OPT_COVER, 1);
+         if (optarg) {
```
nickg, line 255: Can you split this into a function by itself?

```c
cover_mask_t parse_cover_mask(const char *str)
```

src/nvc.c
```diff
@@ -251,7 +252,42 @@ static int elaborate(int argc, char **argv)
...
+                     diag_hint(d, NULL, "selected coverage types shall be "
+                                        "comma separated e.g $bold$--cover=s,t,b$$");
+                     diag_emit(d);
+                     fatal_exit(EXIT_FAILURE);
+                  }
+                  n_chars = 0;
+               }
+               n_chars++;
+               if (n_chars >= 3)
+                  fatal("Invalid coverage type: '%s'. Valid coverage shall be a "
```
nickg, line 283: Just `"invalid coverage type '%'"` should be enough.

nickg (commented):
Please fix those two small things above and then it should be ok

src/nvc.c
Blebowski: Done

src/nvc.c
Blebowski: Done.

Blebowski: But looking at it now, still if we specify explicitly the coverage types, the last one gets skipped. Still need to fix this

nickg:
Sorry could you change one more thing please: the constant `VCODE_VERSION` in `vcode.c` needs to be increment because some of the instructions changed in a backwards-incompatible way.

Blebowski:
> Sorry could you change one more thing please: the constant `VCODE_VERSION` in `vcode.c` needs to be increment because some of the instructions changed in a backwards-incompatible way.

Hi, no problem, will do that. I am currently trying to debug the report generation on "real" project, and I can see its not
optimal the way statements / branches are reported for conditional / selected signal assign. Plus, I ran accross few bugs remaining there, so I am fixing those :)

Blebowski:
@nickg , now the MR is ready from my side. Testing on larger design showed my foolishness in the code which reallocs the chains. I fixed it in little hackish way (instead of using double pointer), but for now it works. Rest is up to some further refactoring, but I would preffer getting it to master. Maybe some other users would start using it, and reporting bugs.

src/parse.c
```diff
@@ -8424,6 +8424,9 @@ static void p_waveform(tree_t stmt, type_t constraint)
 
    while (optional(tCOMMA))
       tree_add_waveform(stmt, p_waveform_element(constraint));
+
+   if (set_loc)
```
nickg, line 8428: Can't this be done unconditionally? The only other place this is called is from the sequential signal assignment statement, which will overwrite the location anyway.

Blebowski: True, I have changed it.

nickg (approved):
Look OK now, thanks.

PR #561, opened 2022-11-05
Remove statement labelling

Blebowski:
Removal of implicit labelling of branches and statements from parser.

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -9260,16 +9260,16 @@ static tree_t p_component_instantiation_statement(ident_t label, tree_t name)
 
    tree_set_loc(t, CURRENT_LOC);
 
-   if (label == NULL) {
+   if (label == NULL)
       parse_error(CURRENT_LOC, "component instantiation statement must "
                   "have a label");
-      tree_set_ident(t, get_implicit_label(t, nametab));
-   }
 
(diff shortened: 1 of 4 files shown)


src/parse.c
```diff
@@ -9263,16 +9260,16 @@ static tree_t p_component_instantiation_statement(ident_t label, tree_t name)
 
    tree_set_loc(t, CURRENT_LOC);
 
-   if (label == NULL) {
+   if (label == NULL)
       parse_error(CURRENT_LOC, "component instantiation statement must "
                   "have a label");
-      tree_set_ident(t, get_implicit_label(t, nametab));
```
nickg, line 9269: Change this to `tree_set_ident(t, error_marker())` which will avoid any crashes later on if some code assumes it has a label.

Blebowski: Done

nickg:
Some of your commits have the email address Ondrej Ille <ondrej.ille@tropicsqaure.com>. Is that correct? You might want to add it to your GitHub account if it is.

Blebowski:
True, thats my work email, and I was working on my PC both for personal stuff, as well as commiting to some of our internal repos where our Gitlab authenticates via LDAP based on email...
But I should probably add it.

Issue #562, opened 2022-11-09
Coverage: Toggle coverage enhancements


Blebowski:
1. [x] Add option to count X,Z -> 0/1 as valid transitions.
2. [x] Add configurable limit on size of signals for which toggle coverage is measured (to reduce performance impact).

Blebowski:
Development in: https://github.com/Blebowski/nvc/tree/coverage-toggle-improve

Blebowski:
Closes #562

Issue #533, opened 2022-09-08
FST dumping - Improvements


Blebowski:
Hi,

I have few remarks for improvements of waveform dumping, IMHO it would be good to include following:
  1. Option to enable/disable dumping of internal signals (not only all signals)
  2. Option to enable/disable dumping of instance ports
  3. Add dumping of variables within processes and shared variables, option to enable/disable such dump.
  4. Add dumping of constants, option to enable/disable such dump.

Would it be feasible only in NVC, or does it require extensions of FST API (and therefore GTKWave modifications)?

nickg:
The first two should be possible. But I don't think FST or GtkWave has any concept of variables. I'm also not sure when you would dump them: at the end of each time step? There also isn't a notion of `'EVENT` for variables so it's not obvious when a value change should be emitted.

Blebowski:
Well there are two options I think:
 - At the end of each non-delta cycle
 - Each time when variable is assigned (which can be multiple times during single simulation time).

AFAIK, GTKWave does not support viewing / expanding delta-cycles on dumped objects, which would be also nice feature, but it would require changes in GTKWave, as well as dumping also information about delta-cycle. This is usefull when debugging heavily clock gated designs which can have lot of delta-cycle "ghost" issues...

I think each time when variable is assigned would be best. Constants obviously would need to be dumped only during initial assignment.

nickg:
> AFAIK, GTKWave does not support viewing / expanding delta-cycles on dumped objects, which would be also nice feature, but it would require changes in GTKWave, as well as dumping also information about delta-cycle. 

I also think this would be a useful feature, and the implementation on the NVC side would be trivial - just need to change the callback from postponed to non-postponed and emit the delta cycle number somewhere. But I don't know how much effort it will be on the FST/GtkWave side.

> I think each time when variable is assigned would be best. Constants obviously would need to be dumped only during initial assignment.

The overhead of dumping variables on every assignment would be extreme - not just because of the data generated (e.g. in loops) but to dump the new value would require a runtime call after every assignment which will basically inhibit almost all optimisations LLVM does now. Dumping all the variables on the last delta cycle is probably ok (although still need to be careful e.g. with large arrays). 

I actually think this sort of thing would be better in some interactive debug environment where you could pause the simulation, inspect variable values, single step, etc.

Blebowski:
OK, I see, then dumping at the end of non-delta cycle seems more reasonable.

[...]

PR #576, opened 2022-11-18
Coverage toggle improve.

Blebowski:
PR with following changes:
 - Fix incorrect reporting of toggle coverage on record elements (keep track of hierarchy level to distinguish hierarchy from scope)
 - Add toggle coverage emit on nested arrays.
 - Fix incorrect numbering of cover tags for multi-dimensional arrays, take into account current dimension as well as TO/DOWNTO range of array.
 - Rename coverage collection options to full words.
 - Add options to exclude multidimensional arrays and arrays of certain size from toggle coverage.
 - Add option to count 'U' -> 0 / 1 as 0/1->1/0 toggles. Add option to count X->1/0 as 0/1 -> 1/0 and 1/0 -> X as 1/0 -> 0/1.


Blebowski:
Closes #562

src/opt.c
```diff
@@ -103,6 +104,71 @@ bool opt_get_verbose(opt_name_t name, const char *filter)
...
+            }
+            // Check for regular options
+            else if (!strcmp(curr_opt->opt, buf)) {
+               break;
+            }
+            if (i == allowed_cnt - 1)
+               curr_opt = NULL;
+         }
+
+         if ((curr_opt == NULL) || (len == ARRAY_LEN(buf))) {
```
nickg, line 141: Needs to be `ARRAY_LEN(buf) - 1` or the string won't be properly null-terminated. But see the comment above about removing `buf`.

src/opt.c
```diff
@@ -103,6 +104,71 @@ bool opt_get_verbose(opt_name_t name, const char *filter)
       return strstr(filter, value) != NULL;
 }
 
+int opt_parse_comma_separated(const char *opt, const char *optarg,
+                              opt_separed_t *allowed, int allowed_cnt,
+                              char **wc_buf)
+{
+   assert(allowed_cnt > 1);
+
+   char buf[64] = {0};
```
nickg, line 113: You don't need to copy the option to a separate buffer. Just keep a pointer to the start of the option name and then use `strncmp`. Something like:

```c
   int rv = 0;
   const char *start = optarg;

   while (1) {

      // Process buffer gathered so far, search if any allowed options match
      if (*optarg == ',' || *optarg == '\0') {
         opt_separed_t *curr_opt;
         for (int i = 0; i < allowed_cnt; i++) {
            curr_opt = &(allowed[i]);
            const char *wc = strchr(curr_opt->opt, '*');
            int wc_pos =  wc - curr_opt->opt;

            // Check for wild-card options -> Copy matched end to wildcard buffer
            if (wc && !strncmp(curr_opt->opt, start, wc_pos)) {
               strcpy(*wc_buf, start + wc_pos);
               wc_buf++;
               break;
            }
            // Check for regular options
            else if (!strncmp(curr_opt->opt, start, optarg - start)) {
               break;
            }
            if (i == allowed_cnt - 1)
               curr_opt = NULL;
         }

         if (curr_opt == NULL) {
            /*  .... */
         } else
            rv |= curr_opt->mask;

         if (*optarg == '\0')
            break;

         start = optarg + 1;
      }
      optarg++;
   }
```

src/opt.c
```diff
@@ -103,6 +104,71 @@ bool opt_get_verbose(opt_name_t name, const char *filter)
...
+      if (*optarg == ',' || *optarg == '\0') {
+         opt_separed_t *curr_opt;
+         for (int i = 0; i < allowed_cnt; i++) {
+            curr_opt = &(allowed[i]);
+            const char *wc = strchr(curr_opt->opt, '*');
+            int wc_pos =  wc - curr_opt->opt;
+
+            // Check for wild-card options -> Copy matched end to wildcard buffer
+            if (wc && !strncmp(curr_opt->opt, buf, wc_pos)) {
+               strcpy(*wc_buf, buf + wc_pos);
```
nickg, line 129: What happens when the option string is longer than the buffer at `*wc_buf`?

```
$ ./bin/nvc -e cover_stmt --cover=ignore-arrays-from-xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
malloc(): invalid size (unsorted)
malloc(): invalid size (unsorted)
```

src/opt.c
```diff
@@ -103,6 +104,71 @@ bool opt_get_verbose(opt_name_t name, const char *filter)
       return strstr(filter, value) != NULL;
 }
 
+int opt_parse_comma_separated(const char *opt, const char *optarg,
```
nickg, line 107: There's really no need to make a separate generic function for this. It would be much simpler, easier to understand, and more correct if there was just a single function in `nvc.c` where we have `parse_cover_mask` now:

```c
void parse_cover_options(cover_mask_t *mask, int *array_limit);
```

That would completely avoid the need for the special "wildcard" handling as you could parse `ignore-arrays-from` specially.

src/lower.c
```diff
@@ -1525,31 +1526,70 @@ static int32_t lower_toggle_tag_for(type_t type, tree_t where, ident_t prefix, i
...
+      int32_t first_tag = -1;
       int64_t low, high;
 
       if (folded_bounds(r, &low, &high)) {
          assert(low <= high);
-         for (int64_t i = low; i <= high; i++) {
+
+         int64_t i;
+         int64_t last;
+         int64_t first;
```
nickg, line 1539: Put these all on one line

src/lower.c
```diff
@@ -1525,31 +1526,70 @@ static int32_t lower_toggle_tag_for(type_t type, tree_t where, ident_t prefix, i
...
+         int inc;
+
+         if (cover_skip_array_toggle(cover_tags, high - low + 1))
+            return -1;
+         cover_add_dim(cover_tags);
```
nickg, line 1544: I'm not sure what `cover_add_dim` / `cover_sub_dim` do exactly. Does it just end up the same as the `curr_dim` argument to this function?

src/nvc.c
```diff
@@ -291,10 +252,27 @@ static int elaborate(int argc, char **argv)
          opt_set_str(OPT_DUMP_VCODE, optarg ?: "");
          break;
       case 'c':
-         if (optarg)
-            cover_mask = parse_cover_mask(optarg);
+         if (optarg) {
+            char LOCAL *wc_buf[1] = {
```
nickg, line 256: This actually does the equivalent of `free(wc_buf)` instead of `free(wc_buf[0])` which is presumably what you want. But see my comment above: this code would be much less convoluted if we just had a single dedicated function for parsing the coverage options.

src/nvc.c
```diff
@@ -291,10 +252,27 @@ static int elaborate(int argc, char **argv)
...
+               xcalloc(32)
+            };
+            opt_separed_t cov_opts[] = {
+               {.opt = "statement",             .mask = COVER_MASK_STMT},
```
nickg, line 260: No need to name the fields explicitly: there's only two fields and they're assigned in order anyway.

```c
{ "statement",    COVER_MASK_STMT },
```

src/nvc.c
```diff
@@ -291,10 +252,27 @@ static int elaborate(int argc, char **argv)
          opt_set_str(OPT_DUMP_VCODE, optarg ?: "");
          break;
       case 'c':
-         if (optarg)
-            cover_mask = parse_cover_mask(optarg);
+         if (optarg) {
+            char LOCAL *wc_buf[1] = {
+               xcalloc(32)
```
nickg, line 257: There's nothing stopping the user putting more than 32 characters after `ignore-arrays-from-`.

src/rt/cover.h
```diff
@@ -106,27 +109,37 @@ typedef enum {
...
-   COVER_MASK_STMT   = (1 << 0),
-   COVER_MASK_BRANCH = (1 << 1),
-   COVER_MASK_TOGGLE = (1 << 2),
+   COVER_MASK_STMT                        = (1 << 0),
+   COVER_MASK_BRANCH                      = (1 << 1),
+   COVER_MASK_TOGGLE                      = (1 << 2),
+   COVER_MASK_TOGGLE_COUNT_FROM_UNDEFINED = (1 << 8),
+   COVER_MASK_TOGGLE_COUNT_FROM_TO_Z      = (1 << 9),
+   COVER_MASK_TOGGLE_IGNORE_MEMS          = (1 << 10),
+   COVER_MASK_TOGGLE_IGNORE_ARRAYS_FROM   = (1 << 11)
```
nickg, line 118: This isn't really a coverage type though is it? It seems `COVER_MASK_TOGGLE_IGNORE_ARRAYS_FROM` is only added here because of the constraints of the comma separated option parsing interface.

src/rt/cover.h
```diff
@@ -106,27 +109,37 @@ typedef enum {
...
 } cover_mask_t;
 
-#define COVER_MASK_ALL (COVER_MASK_STMT | COVER_MASK_BRANCH | COVER_MASK_TOGGLE)
+#define COVER_MASK_ALL_TYPES (COVER_MASK_STMT | COVER_MASK_BRANCH | COVER_MASK_TOGGLE)
```
nickg, line 121: Why change this? It seems to me that `COVER_MASK_ALL` is both shorter and clearer (the other masks don't have `TYPE` in them).

src/rt/cover.h
```diff
@@ -106,27 +109,37 @@ typedef enum {
...
+
+typedef struct {
+   cover_mask_t   mask;
+   int            array_limit;
+} cover_opts_t;
 
-cover_tagging_t *cover_tags_init(cover_mask_t mask);
+cover_tagging_t *cover_tags_init(cover_opts_t *opts);
```
nickg, line 128: No need to pass a structure of options: just pass two separate arguments.

src/rt/model.c
```diff
@@ -1802,6 +1802,11 @@ static void emit_coverage(rt_model_t *m)
    }
 }
 
+cover_tagging_t *get_rt_coverage(rt_model_t *m)
```
nickg, line 1805: Name it `get_coverage`.

test/regress/testlist.txt
```diff
@@ -260,7 +260,9 @@ issue111        normal
 issue115        normal
 issue116        normal
 issue112        normal
-cover1          cover,shell
+cover_stmt      cover,shell
+cover_toggle    cover,shell
+cover_branch    cover,shell
```
nickg, line 265: Please follow the existing convention where tests are named `<topic><number>` so these new tests should be named `cover2` and `cover3`. New tests go at the end of the file.

test/test_lower.c
```diff
@@ -2008,7 +2008,9 @@ START_TEST(test_cover)
 
    tree_t e = run_elab();
 
-   cover_tagging_t *tagging = cover_tags_init(COVER_MASK_ALL);
+   cover_opts_t cov_opts;
+   cov_opts.mask = COVER_MASK_ALL_TYPES;
```
nickg, line 2012: This leaves `cov_opts.array_limit` uninitialised. Use a C99-style designated initialiser instead which zeros out the other members:

```c
cover_opts_t cov_opts = { .mask = COVER_MASK_ALL_TYPES };
```

But I'd rather we just pass two separate arguments.

nvc.1
```diff
@@ -501,6 +505,58 @@ hierarchy coverage report in HTML format, run:
 .Bd -literal -offset indent
 $nvc -c --merge=merged.covdb --report=<path_to_folder_for_html_report> first.covdb second.covdb third.covdb ...
 .Ed
```
nickg, line 507: Start a new paragraph after the `.Ed` with `.Pp` on a line by itself.

nvc.1
```diff
@@ -501,6 +505,58 @@ hierarchy coverage report in HTML format, run:
...
+.It
+.Cm ignore-mems
+- When set, NVC does not collect toggle coverage on multidimensional arrays or
+nested arrays (array of array).
+.It
+.Cm ignore-arrays-from-<size>
+- When set, NVC does not collect toggle coverage for any array which is equal to
+or larger than
+.Cm <size>
+.El
```
nickg, line 553: Add a `.Pp` after `.El` too.

nvc.1
```diff
@@ -501,6 +505,58 @@ hierarchy coverage report in HTML format, run:
 .Bd -literal -offset indent
 $nvc -c --merge=merged.covdb --report=<path_to_folder_for_html_report> first.covdb second.covdb third.covdb ...
```
nickg, line 506: This was from the previous patch but this line is really long and wraps if you view it with `man` in a terminal. Break it over two lines like:

```
.Bd -literal -offset indent
$ nvc -c --merge=merged.covdb --report=<path_to_folder_for_html_report> \\
        first.covdb second.covdb third.covdb ...
.Ed
```

nvc.1
```diff
@@ -501,6 +505,58 @@ hierarchy coverage report in HTML format, run:
...
+- When set, NVC also counts toggles
+.Cm U
+->
+.Cm 1
+as
+.Cm 0
+->
+.Cm 1
+and toggles
+.Cm U
```
nickg, line 521: I would probably just write it plainly like `U -> 0`. I'm not sure the highlighting adds much.

nickg (changes requested):

src/opt.c
Blebowski: I would keep this one generic, since I think it can be usefull for other coverage unrelated options (thats why I changed it to be generic). E.g. from: https://github.com/nickg/nvc/issues/533

I can imagine something like:
`--dump-objs=signals,ports,memories,psl,constants,variables`

src/lower.c
Blebowski: Not really, `curr_dim` tracks current dimension in case of multi-dimensional array.
Dimension within `cover_tags`, tracks also dimension in case of nested array. Dimension
from nested array is not really held anywhere, since that is "array of array of array of...".
So, I need to track the information that I am currently already in second level.

Blebowski: So that I can properly detect something as in `cover2` test.

src/nvc.c
Blebowski: I reworked the `wc_buf` to be part of `opt_separed_t`. This fixes the issue, leaves no leak, and
if there are multiple wild-card commands, its easy to find wild-card for each command.

src/nvc.c
Blebowski: I added protection into `opt_parse_comma_separated`.

src/rt/cover.h
Blebowski: Yes, does it mind?

But still the information "exclude / not exclude arrays" still needs to be stored somewhere.

src/rt/cover.h
Blebowski: I changed it back. The idea was to say by `TYPE` that this is mask with all coverage types, not necessarily
all coverage options.

src/rt/cover.h
Blebowski: Done

test/regress/testlist.txt
Blebowski: Done

test/test_lower.c
Blebowski: Done.

nvc.1
Blebowski: Done.

nvc.1
Blebowski: Done.

nvc.1
Blebowski: I would keep it like this. I see that in `man` in command line, its not really important, but in HTML format the highlighted version seems more readable.

src/opt.c
nickg: I would just write a simple non-generic function like this:

```c
static void parse_cover_options(const char *str, cover_mask_t *mask,
                                int *array_limit)
{
   static const struct {
      const char *name;
      cover_mask_t mask;
   } options[] = {
      { "statement",             COVER_MASK_STMT                        },
      { "toggle",                COVER_MASK_TOGGLE                      },
      { "branch",                COVER_MASK_BRANCH                      },
      { "all",                   COVER_MASK_ALL                         },
      { "count-from-undefined",  COVER_MASK_TOGGLE_COUNT_FROM_UNDEFINED },
      { "count-from-to-z",       COVER_MASK_TOGGLE_COUNT_FROM_TO_Z      },
      { "ignore-mems",           COVER_MASK_TOGGLE_IGNORE_MEMS          },
   };

   for (const char *start = str; ; str++){
      if (*str == ',' || *str == '\0') {
         if (strncmp(start, "ignore-arrays-from-", 19) == 0)
            *array_limit = atoi(start + 19);
         else {
            int pos = 0;
            for (; pos < ARRAY_LEN(options); pos++) {
               if (!strncmp(options[pos].name, start, str - start))
                  break;
            }

            if (pos == ARRAY_LEN(options)) {
               diag_t *d = diag_new(DIAG_FATAL, NULL);
               diag_printf(d, "unknown coverage type '%.*s'",
                           (int)(str - start), start);
               diag_hint(d, NULL, "valid coverage types are: \n"
                         "  statement\n"
                         "  toggle\n"
                         "  branch");
               diag_hint(d, NULL, "selected coverage types shall be "
                         "comma separated e.g $bold$--cover=toggle,branch$$");
               diag_emit(d);
               fatal_exit(EXIT_FAILURE);
            }
            else
               *mask |= options[pos].mask;
         }

         if (*str == '\0')
            break;

         start = str + 1;
      }
   }
}
```

This is clear and easy to read, and also has a more precise error message. If we need comma separated argument parsing elsewhere I would rather just implement it separately. The problem with making a generic function for this is that it will become more and more complicated as we add new special cases to handle different types of options (like the wildcard parsing here). I would much rather have two simple functions that duplicate some logic than one complicated/obfuscated shared function.

nickg:
BTW if you're going to reply to a lot of comments at once, you can use the "start a review" button. That will send one notification rather than N.

Blebowski:
OK, I will use the review button. I did not know it spams you with bunch of emails :O

src/opt.c
Blebowski: OK then, I will add the specific function, but TBH in this case I think having generic function
would be better than duplicated logic (as long as we don't bloat it with 10 other features).
I understand that sometimes bit of duplication is better than bit of dependency, but I am
afraid to end up with couple of different option parsers, one for each command, and have
different behavior in seemingly equal situations.

E.g. if I look at VCS parsing of command line arguments, it differs very much for VHDL (`vhdlan`) and
for Verilog (`vlogan`) compilation commads. So e.g. passing `std_logic_vector` to VHDL generic, and
passing `logic[X:Y]` as Verilog parameter requires completely different command line options, different
way how to escape the values, etc... The resulting user experience is __pain__.

It originates from the fact that those were two different compilers long time ago, but still, I would like to avoid this kind of situation.

Blebowski:
I went through the issues you pointed out, and also added tests for most of the new features.

src/nvc.c
```diff
@@ -219,43 +219,56 @@ static void set_top_level(char **argv, int next_cmd)
...
+      { "statement",             COVER_MASK_STMT                        },
+      { "toggle",                COVER_MASK_TOGGLE                      },
+      { "branch",                COVER_MASK_BRANCH                      },
+      { "all",                   COVER_MASK_ALL                         },
+      { "count-from-undefined",  COVER_MASK_TOGGLE_COUNT_FROM_UNDEFINED },
+      { "count-from-to-z",       COVER_MASK_TOGGLE_COUNT_FROM_TO_Z      },
+      { "ignore-mems",           COVER_MASK_TOGGLE_IGNORE_MEMS          },
+   };
+
+   for (const char *start = str; ; str++){
```
nickg, line 238: Missing a space before the { (I think that was my mistake).

src/nvc.c
```diff
@@ -269,7 +282,7 @@ static int elaborate(int argc, char **argv)
       { 0, 0, 0, 0 }
    };
 
-   cover_mask_t cover_mask = 0;
+   cover_opts_t cover_opts = {0};
```
nickg, line 285: No need to have the struct, just use two variables

```c
cover_mask_t cover_mask = 0;
int cover_array_limit = 0;
```

src/opt.c
```diff
@@ -17,6 +17,7 @@
 
 #include "opt.h"
 #include "util.h"
+#include "diag.h"
```
nickg, line 20: Not needed anymore

src/opt.h
```diff
@@ -64,7 +64,6 @@ void opt_set_str(opt_name_t name, const char *val);
 int opt_get_int(opt_name_t name);
 const char *opt_get_str(opt_name_t name);
 bool opt_get_verbose(opt_name_t name, const char *filter);
-
```
nickg, line 67: Restore the blank line

src/rt/cover.c
```diff
@@ -296,17 +325,19 @@ void cover_dump_tags(cover_tagging_t *ctx, fbuf_t *f, cover_dump_t dt,
    ident_write_end(ident_ctx);
 }
 
-cover_tagging_t *cover_tags_init(cover_mask_t mask)
+cover_tagging_t *cover_tags_init(cover_mask_t mask, int array_limit)
 {
    cover_tagging_t *ctx = xcalloc(sizeof(cover_tagging_t));
-   ctx->mask = mask;
+   ctx->opts.mask = mask;
+   ctx->opts.array_limit = array_limit;
```
nickg, line 332: Don't need the struct, just have two separate fields.

src/rt/cover.c
```diff
@@ -162,6 +164,23 @@ bool cover_is_stmt(tree_t t)
    }
 }
 
+bool cover_skip_array_toggle(cover_tagging_t *tagging, int a_size)
+{
+   assert (tagging);
+
+   // Array is equal to or than configured limit
+   if (cover_enabled(tagging, COVER_MASK_TOGGLE_IGNORE_ARRAYS_FROM) &&
```
nickg, line 172: This could be `if (tagging->array_limit != 0 && a_size >= tagging->array_limit)` and then you don't need this enum value.

test/test_lower.c
```diff
@@ -2008,7 +2008,8 @@ START_TEST(test_cover)
 
    tree_t e = run_elab();
 
-   cover_tagging_t *tagging = cover_tags_init(COVER_MASK_ALL);
+   cover_opts_t cov_opts = {.mask = COVER_MASK_ALL};
+   cover_tagging_t *tagging = cover_tags_init(cov_opts.mask, cov_opts.array_limit);
```
nickg, line 2012: `cover_tags_init(COVER_MASK_ALL, 0)`

src/lower.c
nickg: OK, but it's confusing that "dimension" is used to refer to both the actual array dimension and the nesting level of arrays in arrays. Can you call the later "depth" or "nesting" or something like that? E.g. `cover_inc_nesting()`, `cover_dec_nesting()`.

Blebowski:
Hi, I fixed all the remarks you had. Looking at the coverage of the cover code its better now, but there are still some holes. I will extend the existing tests, and/or add new ones to cover these some time soon.

nickg (approved):
It would be useful for testing if we had a simple text report that looked something like:

```
Instance    Statement    Branch  Toggle
A.B         5/12         1/2     5/7
A.B.C       1/7          3/6     6/6
```

That would be a bit more accurate than just checking the overall percentages as we do at the moment. But it's not necessary to do that now.

PR #578, opened 2022-12-03
src: Expression coverage

Blebowski:
This is first draft of added expression coverage. The MR implements:

1. Expression coverage for different types of expressions (AND, OR, =, /=, etc...)
2. Naming refactoring inside the coverage code to shorten the enum values and keep the code readable.
3. Usage of defined enums on more places rather than hard-coded values.
4. Reworked Coverage report to keep together tags related to single tree item. Different flags for single tag are kept in table with "bins", report is reworked to be better readable.

Currently, the implementation is not yet final, but ready for first review. I still would like to add two things in this MR:
1. Add option to print per-hierarchy coverage (as you already suggested @nickg previously), to have better error checking mechanism in NVCs regression.
2. Add Expression coverage on logic ops with std_logic (e.g. to treat `std_log_a_sig OR std_log_b_sig` the same way as in LHS/RHS are booleans).


src/lower.c
```diff
@@ -4344,42 +4499,67 @@ static vcode_reg_t lower_expr(tree_t expr, expr_ctx_t ctx)
 {
    PUSH_DEBUG_INFO(expr);
 
+   if (cover_enabled(cover_tags, COVER_MASK_EXPRESSION))
+      cover_push_scope(cover_tags, expr);
```
nickg, line 4503: Why do you need to create a separate scope for each expression? I think the overhead of this will be pretty large.

nickg:
> Currently, the implementation is not yet final, but ready for first review. I still would like to add two things in this MR:

Please split these into separate PRs: it's easier to review small changes.

Blebowski:
Hello @nickg,

I have modified the expression coverage not to push coverage scope with each expression (It was indeed overkill). Now scopes are pushed only for T_FCALL.

I have split the work into stand-alone commits for better readability / reviewability, but its still not 100%
independent. The commits are not 100% independent though, so it hard to split them into multiple MRs.
I did not add the "hierarchy dump" option, I will do so in the next PR.

The tests of coverage will soon need the "hierarchy dump" since right now there are places where
double errors (e.g. accidentally inverted `=` and `/=` will make test pass without noticing), so that
will be probably next step.

I decided to first finish the expression coverage and optimize the merging only then, as its better
to first have some more-less feature complete implementation, and optimize only afterwards.

As for the expression coverage, I do most of the work on VCODE level, I emit operations
needed to build up the mask for coverage bins, and CGEN level then only ORs the mask with
previous value. IMHO it is better since you plan to move to JIT IIR, and thus keeping most
of coverage measurement logic on VCODE will avoid complete re-work when JIT IIR is introduced.
That being said, current expression coverage is not optimal, and I would welcome suggestions
how to improve it.

src/lower.c
```diff
@@ -1663,20 +1656,163 @@ static void lower_toggle_coverage(tree_t decl)
...
+         {COV_FLAG_10, lhs,   rhs_n},
+         {COV_FLAG_11, lhs,   rhs},
+      };
+
+      // Check LHS/RHS combinations
+      for (int i = 0; i < ARRAY_LEN(bins); i++)
+         if (flags & bins[i].flag) {
+            vcode_reg_t tmp = emit_and(bins[i].lhs, bins[i].rhs);
+            vcode_reg_t set_bit = emit_select(tmp, emit_const(vc_int, bins[i].flag),
+                                                   emit_const(vc_int, 0));
```
nickg, line 1729: `vcode_reg_t zero = emit_const(vc_int);` before line 1725 and then use it here will be slightly more efficient.

src/lower.c
```diff
@@ -1663,20 +1656,163 @@ static void lower_toggle_coverage(tree_t decl)
...
+            mask = emit_add(mask, set_bit);
+         }
+
+      lower_expression_coverage(fcall, flags, mask);
+   }
 
    return result;
 }
 
+static void lower_std_logic_expr_coverage(tree_t fcall, tree_t decl,
```
nickg, line 1739: I think you can split this up into two separate PRs: one to add the basic expression coverage functionality, and a second one to extend it to std_logic.

src/lower.c
```diff
@@ -1663,20 +1656,163 @@ static void lower_toggle_coverage(tree_t decl)
...
+   if (type_is_array(type_param(tree_type(decl), 0)))
+      return;
+
+   flags |= COV_FLAG_EXPR_STD_LOGIC;
+
+   vcode_reg_t lhs = args[1];
+   vcode_reg_t rhs = args[2];
+
+   // TODO: Reference std_ulogic type lowered values somehow, do not
+   //       hard-code as we expect to get it translated.
```
nickg, line 1783: You can just do `vcode_type_t logic_type = vcode_reg_type(args[1]);`

src/lower.c
```diff
@@ -2278,9 +2422,16 @@ static vcode_reg_t lower_fcall(tree_t fcall)
 {
    tree_t decl = tree_ref(fcall);
 
+   if (cover_enabled(cover_tags, COVER_MASK_EXPRESSION))
```
nickg, line 2425: I think these scopes will only ever have two tags in them (perhaps we should call them coverage bins instead). Could we instead just number expressions sequentially inside the parent scope?

src/rt/cover.c
```diff
@@ -165,6 +170,24 @@ bool cover_is_stmt(tree_t t)
    }
 }
 
+static bool cover_is_branch(tree_t branch)
+{
+   if (tree_kind(branch) == T_ASSOC || tree_kind(branch) == T_COND)
```
nickg, line 175:
```suggestion
   return tree_kind(branch) == T_ASSOC || tree_kind(branch) == T_COND;
```

src/vcode.h
```diff
@@ -105,6 +105,7 @@ typedef enum {
    VCODE_OP_COVER_STMT,
    VCODE_OP_COVER_BRANCH,
    VCODE_OP_COVER_TOGGLE,
+   VCODE_OP_COVER_EXPRESSION,
```
nickg, line 108: Call it `VCODE_OP_COVER_EXPR` instead otherwise the long name is too unwieldy.

nickg (changes requested):

src/lower.c
Blebowski: Yes, that is a good idea. The original idea of "scope for everything" which targeted independce
of coverage hierarchy between different parts of large `case` statements representing complex
FSM falls apart here. If someone would have way too complex / nested logic expression
representing some combo logic, its anyway good coding practice to create sub-signals and
split such expression to multiple stand-alone expressions, each nested under separate
signal assignment.

Blebowski:
I have processed your comments, and split the expression coverage on std_logic operands
to separate MR.

Blebowski:
The expression coverage passes, but still crashes my IP project, and also some testing projects from vhdl_tests (neorv32), so I still need to dig into it. Maybe something related to incorrectly emited code for mask creation on VCODE.

src/vcode.h
```diff
@@ -105,6 +105,7 @@ typedef enum {
    VCODE_OP_COVER_STMT,
    VCODE_OP_COVER_BRANCH,
    VCODE_OP_COVER_TOGGLE,
+   VCODE_OP_COVER_EXPR,
```
nickg, line 108: You need to increment `VCODE_VERSION` every time you make an incompatible change otherwise user will get confusing errors when loading existing files (adding an new opcode to the middle of the list changes all subsequent ones).

test/regress/testlist.txt
```diff
@@ -698,3 +698,4 @@ cover4          cover=branch,gold
 cover5          cover,shell
 cover6          cover,shell
 issue577        normal,2008
+cover8          cover,shell
```
nickg, line 701:
```
cover8       cover=expression,gold
```

You don't need the shell script any more, see f03ae9c616

nvc.1
```diff
@@ -485,11 +485,46 @@ branch (one for each of true/false)
 .Ql std_logic
 (including nested arrays) creates two coverage bins (to track
 \fB0\fP -> \fB1\fP and \fB1\fP -> \fB0\fR transitions).
+.It
+.Cm expression
```
nickg, line 489:
```suggestion
.Cm expression
- NVC creates multiple coverage bins for combinations of input operands
of the following logical operators:
.Ql and Ns ,
.Ql nand Ns ,
.Ql or Ns ,
.Ql nor Ns ,
.Ql xor Ns ,
.Ql xnor Ns ,
such that propagation of operand values causes the expression result to
change its value.  Further, NVC creates two coverage bins for evaluating
expression result to
.Ql True
and
.Ql False
for the following operators:
.Ql = Ns ,
.Ql /= Ns ,
.Ql > Ns ,
.Ql < Ns ,
.Ql <= Ns ,
.Ql >= Ns ,
.Ql not Ns .
```

src/lower.c
```diff
@@ -1663,16 +1656,83 @@ static void lower_toggle_coverage(tree_t decl)
...
 {
-   // TODO: Change to expression coverage, not branch!
-   //
-   //int32_t cover_tag, sub_cond;
-   //if (!cover_is_tagged(cover_tags, fcall, &cover_tag, &sub_cond))
-   //   return result;
-   //
-   //if (sub_cond > 0)
-   //   emit_cover_cond(result, cover_tag, sub_cond);
+   if (cover_enabled(cover_tags, COVER_MASK_EXPRESSION)) {
```
nickg, line 1673:
```suggestion
   if (!cover_enabled(cover_tags, COVER_MASK_EXPRESSION))
      return;
```

nickg:
> The expression coverage passes, but still crashes my IP project, and also some testing projects from vhdl_tests (neorv32), so I still need to dig into it. Maybe something related to incorrectly emited code for mask creation on VCODE.

It generates invalid code in cases like:

```vhdl
if A and B then ...
```

If evaluating `B` might have side effects (e.g. call to non-builtin function, division by zero, etc.) then `B` is only evaluated if `A` is false but `lower_logical` is using both results regardless.

Blebowski:
> > The expression coverage passes, but still crashes my IP project, and also some testing projects from vhdl_tests (neorv32), so I still need to dig into it. Maybe something related to incorrectly emited code for mask creation on VCODE.
> 
> It generates invalid code in cases like:
> 
> ```vhdl
> if A and B then ...
> ```
> 
> If evaluating `B` might have side effects (e.g. call to non-builtin function, division by zero, etc.) then `B` is only evaluated if `A` is false but `lower_logical` is using both results regardless.

I see, I was wondering what the LLVMs "Instruction does not dominate all uses" means, but now it makes sense. The question is, how to fix it in such way, that I still get the expression bins like LHS = 0,  RHS = 0 for `OR` expression coverable. Isn't the only solution to always (when coverage is enabled) execute also RHS argument (which might have performance impact) ?

Blebowski:
> > > The expression coverage passes, but still crashes my IP project, and also some testing projects from vhdl_tests (neorv32), so I still need to dig into it. Maybe something related to incorrectly emited code for mask creation on VCODE.
> > 
> > 
> > It generates invalid code in cases like:
> > ```vhdl
> > if A and B then ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If evaluating `B` might have side effects (e.g. call to non-builtin function, division by zero, etc.) then `B` is only evaluated if `A` is false but `lower_logical` is using both results regardless.
> 
> I see, I was wondering what the LLVMs "Instruction does not dominate all uses" means, but now it makes sense. The question is, how to fix it in such way, that I still get the expression bins like LHS = 0, RHS = 0 for `OR` expression coverable. Isn't the only solution to always (when coverage is enabled) execute also RHS argument (which might have performance impact) ?

But, thinking about it, if second argument is executed regardless of first argument value (when expression coverage is enabled), then there might be simulation mismatch between coverage enabled / disabled cases :/
then simulat

nickg:
> Isn't the only solution to always (when coverage is enabled) execute also RHS argument (which might have performance impact) ?

It's not really about performance: it's not correct to evaluate the RHS if the result can be determined solely from the LHS. The LRM specifies it this way in section 9.2.2 of the 2008 revision:

> For the short-circuit operations and, or, nand, and nor on types BIT and BOOLEAN, the right operand is evaluated only if the value of the left operand is not sufficient to determine the result of the operation.

Consider something like:

```vhdl
if x /= 0 and y / x = 42 then
```

Blebowski:
I see, that makes sense. Just as you wrote it, I made a modification of `cover8` and was trying very similar
divide by zero example.

OK, so for now I would emit the coverage code for the expression only inside the `arg1` block.
This would result in some expressions being uncoverable (e.g.  01 bin on `and` expression).
But, still it makes more sense to add it (then to completely skip it for expressions with side-effect),
as 11 or 10 are still coverable.

But this might be a good match for automatically excluding such bin from expression coverage
via a dedicated option which will automatically exclude uncoverable expressions or their bins.

Blebowski:
I have processed your comments.

nickg (approved):
I removed the now redundant `cover8.sh` and fixed a few minor formatting issues in `lower.c`.

PR #588, opened 2023-01-15
Coverage exclude support

Blebowski:
Hello @nickg,

This is first draft of "Exclude file" support for coverage report generation.

The syntax of exclude file is like so:
`exclude <TAG_HIER> <BIN>`

`#` is treated as comment


src/rt/cover.c
```diff
@@ -776,6 +811,187 @@ void x_cover_setup_toggle_cb(sig_shared_t *ss, int32_t *toggle_mask)
...
+   cover_exclude_ctx_t ctx = {0};
+
+   ctx.ef = fopen(path, "r");
+   if (ctx.ef == NULL)
+      fatal("Failed to open exclude file: %s\n", path);
+
+   char *delim = " ";
+   ssize_t read;
+
+   while ((read = getline(&(ctx.line), &(ctx.line_len), ctx.ef)) != -1) {
```
nickg, line 962: `getline` is missing on Windows but we can just copy the definition from Gnulib. I'll do that later.

Blebowski:
I have no idea why is the Ubuntu test failing. I have checked it on my side, and its passing (and I am developing also on Ubuntu 20.04). Any clues?

Blebowski:
OK, the problem was between keyboard and the chair...

Blebowski:
On my side, the MR is ready for review (apart from the `getline` fix). I have added test to go through the
corner-cases. It is quiet complex, and eventually might be split into multiple tests, but I hope for now
it is sufficient.

Now, the user experience of the exclude feature will be terrible. Because unless building with `COVER_DEBUG`,
one has no chance how to see the `HIERARCHY` which shall be placed to the exclude file. I plan to add a button
with a text like: `Get exclude command` directly into the report (next to each coverage item/bin), that will copy
the exclude command directly to clipboard, and user can then just paste it to opened exclude file, and
re-generate the report. However, I would place it in a separate MR, since you were saying that smaller changes are more easily reviewable.

src/rt/cover.c
nickg: Try rebasing and it should work on Windows

src/rt/cover.c
```diff
@@ -776,6 +814,186 @@ void x_cover_setup_toggle_cb(sig_shared_t *ss, int32_t *toggle_mask)
...
+
+      // Strip comments and newline
```
nickg, line 966: I'd do this directly with a for loop:

```c
char *p = ctx.line;
for (; *p != '\0' && *p != '#' && *p != '\n');
*p = '\0;
```

It's less code and only makes one pass over the array.

src/rt/cover.c
```diff
@@ -776,6 +814,186 @@ void x_cover_setup_toggle_cb(sig_shared_t *ss, int32_t *toggle_mask)
...
+      char *cmnt = strchr(ctx.line, '#');
+      char *nl = strchr(ctx.line, '\n');
+      if (cmnt)
+         *cmnt = '\0';
+      if (nl)
+         *nl = '\0';
+
+      // Parse line as space separated tokens
+      char *tok = strtok(ctx.line, delim);
+      while (tok != NULL) {
```
nickg, line 976:
```suggestion
      for (char *tok = strtok(ctx.line, delim); tok; tok = strtok(NULL, delim)) {
```

nickg (changes requested):

src/rt/cover.c
nickg: Or with a library function you could use `strpbrk`:

```c
char *stop = strpbrk(ctx.line, "#\n");
if (stop)
   *stop = '\0';
```

nickg:
@Blebowski I just pushed one commit to use the same error location reporting as VHDL errors. E.g.

```
** Fatal: invalid bin: 'WTF_BIN' for expression: 'WORK.COVER9._P0._S0._E0'.Valid bins are: 
          BIN_0_0, BIN_0_1, BIN_1_0, BIN_1_1
   > /home/nick/nvc/test/regress/cover9_ef3.txt:3
   |
 3 | exclude WORK.COVER9._P0._S0._E0 WTF_BIN
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Blebowski:
I see, that makes sense. I was wondering exactly about the same thing, that I was duplicating the logic behind
`loc_t`, however, I was not sure how to implement it with it.

nickg:
Yes please fix the conflict in `testlist.txt` and also move the `cover9_ef*.txt` files to a separate directory (add `test/regress/data/`) and then I'll merge it.

nickg:
```
** Fatal: invalid bin: 'WTF_BIN' for expression: 'WORK.COVER9._P0._S0._E0'.Valid bins are: 
          BIN_0_0, BIN_0_1, BIN_1_0, BIN_1_1
```

This message needs a space before "Valid".

nickg:
Oh, I messed up the `gold/cover9.txt` by including the absolute path. I think you can fix that by passing a relative file name in the shell script. E.g. `$(realpath --relative-to=. $TESTDIR/regress/cover9_ef4.txt)`.

Blebowski:
This does not work, nvc figures out the absolute path despite relative one being passed to it. I am trying to replace in flight, but thats lagging on MAC OS.

nickg:
I think '+' is a GNU extension to sed. `s/[^ ]*regress\/data\//data\//g` works ok.

nickg (approved):

PR #600, opened 2023-01-27
src: Add button to code coverage report to get Exclude command.

Blebowski:
This MR adds a button to the generated Code coverage report which copies
the "exclude" command for respective bins / coverage tag into the clipboard.

Code coverage analysis then can work like so:
1. generate report with empty exclude file passed
2. View the report, get the exclude command, paste it to opened exclude file in IDE / Text editor
3. Regenerate report
4. Go to step 2 again until you have reached 100 % or found something what should not be excluded. In such case, write test which stimulates the situation and gets the code covered.

I have seen this sort of "working model" working for many verification engineers.

I tried to keep the changes simple.

The new button has different CSV style than the buttons for switching tabs. I will tackle that in
next MR where I would like to optimize the size of the generated HTML and also speed of generation.

### DIFF
--- a/src/rt/cover.c
+++ b/src/rt/cover.c
@@ -176,6 +176,13 @@ enum std_ulogic {
    _DC = 0x8
 };
 
+typedef enum {
+   PAIR_UNCOVERED    = 0,
+   PAIR_EXCLUDED     = 1,
+   PAIR_COVERED      = 2,
+   PAIR_LAST         = 3
+} cov_pair_kind_t;
+
 bool cover_is_stmt(tree_t t)
 {
    switch (tree_kind(t)) {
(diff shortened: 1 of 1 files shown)


Blebowski:
I did not add a test since I don't really know how to test this feature without some sort-of
"clicking framework" and repetitive check on the generated report. It seemed an overkill
for this case. I have of course tried it manually though.

src/rt/cover.c
```diff
@@ -1301,57 +1308,107 @@ static void cover_print_code_line(FILE *f, loc_t loc, cover_line_t *line)
    fprintf(f, "</code>");
 }
 
-static void cover_print_bins(FILE *f, cover_pair_t *pair)
+static void cover_print_get_exclude_button(FILE *f, cover_tag_t *tag,
+                                           uint32_t flag, bool add_td)
+{
+
+   const char *bin_name = "";
+   if (flag)
```
nickg, line 1316: Put braces around the outer if and for statements

src/rt/cover.c
```diff
@@ -1301,57 +1308,107 @@ static void cover_print_code_line(FILE *f, loc_t loc, cover_line_t *line)
...
+   fprintf(f, "<tr style=\"background-color:#999999;\">");
+   fprintf(f, "<th style=\"width:40px;\"></th>");
+
+   for (int i = 0; i < cols; i++) {
+      const char *val = va_arg(argp, const char *);
+      fprintf(f, "<th style=\"width:100px;\">%s</th>", val);
+   }
+
+   if (pkind == PAIR_UNCOVERED)
+         fprintf(f, "<th style=\"width:150px;\">Exclude Command</th>");
```
nickg, line 1368: Too much indentation

src/rt/cover.c
```diff
@@ -1301,57 +1308,107 @@ static void cover_print_code_line(FILE *f, loc_t loc, cover_line_t *line)
...
+
+   fprintf(f, "</tr>");
+}
+
+static void cover_print_bins(FILE *f, cover_pair_t *pair, cov_pair_kind_t pkind)
 {
    loc_t loc = pair->tag->loc;
+   char t_str[6] = {0};
```
nickg, line 1376: These two don't actually need to be arrays as you only write constant strings into them: just make them `const char *`.

nickg (changes requested):

nickg:
> I did not add a test since I don't really know how to test this feature without some sort-of "clicking framework" and repetitive check on the generated report. It seemed an overkill for this case. I have of course tried it manually though.

It's fine to test this sort of thing manually.

nickg (approved):

PR #602, opened 2023-01-29
Coverage Report Optimizations

Blebowski:
Hello,

this PR improves code coverage report generation:
1. Print only name of the instance in instance sub-table (first commit).
2. Optimize querying name of the signal for toggle coverage:
   - Approach A - Second commit - Dump separately signal name -> Drawbacks: Prolongs intermediate code generation and bloats COVDB size
   - Approach B - Do not track "level" of tag, but track position within `hier` ident where signal / port name starts.

The approach B is better, see the print in the issue related to performance optimizations

### DIFF
--- a/src/rt/cover.c
+++ b/src/rt/cover.c
@@ -219,9 +219,8 @@ static bool cover_is_branch(tree_t branch)
 
 static bool cover_is_toggle_first(tree_t decl)
 {
-   return tree_kind(decl) == T_SIGNAL_DECL ||
-          tree_kind(decl) == T_PORT_DECL;
-
+   const tree_kind_t kind = tree_kind(decl);
+   return kind == T_SIGNAL_DECL || kind == T_PORT_DECL;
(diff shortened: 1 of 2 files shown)


src/rt/cover.c
```diff
@@ -217,6 +217,13 @@ static bool cover_is_branch(tree_t branch)
    return tree_kind(branch) == T_ASSOC || tree_kind(branch) == T_COND;
 }
 
+static bool cover_is_toggle_first(tree_t decl)
+{
+   return tree_kind(decl) == T_SIGNAL_DECL ||
+          tree_kind(decl) == T_PORT_DECL;
+
```
nickg, line 224:
```suggestion
   const tree_kind_t kind = tree_kind(decl);
   return kind == T_SIGNAL_DECL || kind == T_PORT_DECL;
```

nickg (approved):

PR #605, opened 2023-01-30
Code coverage report limit

Blebowski:
Add `--item-limit` option to coverage report generation. Limits number of items printed
for a single type of coverage item. By default 5000. Also adds test for the feature.

In real scenario, user will most likely merge coverage from exhaustive regression run
and then print only `covered` and `excluded` items. However, it is good that by default
NVC does not generate report which causes latest Google Chrome to crash whatever
the state of the design (I was testing with `vhdl_projects` repo and NEORV32 with `./test.sh`
script)


src/nvc.c
```diff
@@ -1098,6 +1100,9 @@ static int coverage(int argc, char **argv)
       case 'd':
          rpt_mask = parse_cover_print_spec(optarg);
          break;
+      case 'l':
+         item_limit = atoi(optarg);
```
nickg, line 1104:
```suggestion
         item_limit = parse_int(optarg);
```

`parse_int` will give an error message for `--item-limit=zzzzz` and similar.

test/regress/gold/cover11.txt
```diff
@@ -0,0 +1,7512 @@
+** Note: excluding toggle: 'WORK.COVER11.ARR_B(1500)' bins: BIN_0_TO_1, BIN_1_TO_0
```
nickg, line 1: This file is too big (and also not used?)

nickg (changes requested):

test/regress/gold/cover11.txt
Blebowski: OK, I removed it and check only that the item is/is not part of the HTML report by grep.

nickg (approved):

PR #611, opened 2023-02-04
WIP: Cover exclude unreachable

Blebowski:
This MR adds basic exclude support on unreachable toggles on instance ports.
Unreachability is remembered in run-time data and checked during report
generation if according CLI option is set. Exclude reason is now placed in
third column of excluded bins to distinguish between the ones coming from
exclude file and automatically excluded bins.

The implementation handles `std_logic` and `std_logic_vector`.

The MR does not handle the cases where part of the port is driven by non-const
value (e.g. single bit of a vector, or single field of a record port).

### DIFF
--- a/src/rt/cover.c
+++ b/src/rt/cover.c
@@ -824,11 +824,11 @@ static bool is_constant_input(rt_signal_t *s)
       decl = sc->where;
    }
 
-   if (tree_kind(decl) != T_PORT_DECL) {
+   if (tree_kind(decl) != T_PORT_DECL)
       return false;
-   } else if (tree_subkind(decl) != PORT_IN) {
+   else if (tree_subkind(decl) != PORT_IN)
       return false;
-   } else {
+   else {
       rt_nexus_t *n = &(s->nexus);
       for (unsigned i = 0; i < s->n_nexus; i++, n = n->chain) {
          if (n->n_sources > 0)
--- a/src/rt/cover.c
+++ b/src/rt/cover.c
@@ -849,8 +849,7 @@ void x_cover_setup_toggle_cb(sig_shared_t *ss, int32_t *toggle_mask)
       for (int i = 0; i < s->shared.size; i++) {
          // Remember un-reachability in run-time data. Exclude mask not
          // available at run-time.
-         (*toggle_mask) |= COV_FLAG_UNREACHABLE;
-         toggle_mask++;
+         (*toggle_mask++) |= COV_FLAG_UNREACHABLE;
       }
       return;
    }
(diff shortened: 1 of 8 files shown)


nickg:
If it's not ready for review yet you can use the "Convert to draft" button the right.

Blebowski:
Well, I think for now the feature is OK to be rewieved / merged.

I guess for the better understanding of how the drivers / nexuses, I need to dig into the `model.c`
and try to run it with the debug prints. I haven't really dived into this part of the code-base, so
that might take time.
 
In future, I can try to figure out how to detect constants driving only parts of the signals, e.g as
in the related issue with `for_sure_toggling` example.

src/rt/cover.c
```diff
@@ -811,13 +811,50 @@ DEFINE_COVER_TOGGLE_CB(cover_toggle_cb_0_1_z,   cover_toggle_check_0_1_z)
...
+         sc = sc->parent;
+      }
+      decl = sc->where;
+   }
+
+   if (tree_kind(decl) != T_PORT_DECL) {
+      return false;
+   } else if (tree_subkind(decl) != PORT_IN) {
+      return false;
+   } else {
```
nickg, line 831:
```suggestion
   if (tree_kind(decl) != T_PORT_DECL)
      return false;
   else if (tree_subkind(decl) != PORT_IN)
      return false;
   else {
```

src/rt/cover.c
```diff
@@ -811,13 +811,50 @@ DEFINE_COVER_TOGGLE_CB(cover_toggle_cb_0_1_z,   cover_toggle_check_0_1_z)
...
    rt_model_t *m = get_model();
    cover_mask_t op_mask = get_coverage(m)->mask;
    sig_event_fn_t fn = &cover_toggle_cb_0_1;
 
+   if (is_constant_input(s)) {
+      for (int i = 0; i < s->shared.size; i++) {
+         // Remember un-reachability in run-time data. Exclude mask not
+         // available at run-time.
+         (*toggle_mask) |= COV_FLAG_UNREACHABLE;
+         toggle_mask++;
```
nickg, line 853:
```suggestion
         (*toggle_mask++) |= COV_FLAG_UNREACHABLE;
```

src/rt/cover.c
```diff
@@ -1664,6 +1715,13 @@ static int cover_append_to_chain(cover_chain_t *chain, cover_tag_t *tag,
    return rv;
 }
 
+static bool cover_tag_unreachable(cover_report_ctx_t *ctx, cover_tag_t *tag)
+{
+   if ((ctx->tagging->mask & COVER_MASK_EXCLUDE_UNREACHABLE) == 0)
+      return false;
+   return (tag->data & COV_FLAG_UNREACHABLE);
```
nickg, line 1722:
```suggestion
   return !!(tag->data & COV_FLAG_UNREACHABLE);
```

Never store or return a `bool` value that's not zero or one. Usually `bool` is typedef-ed to `char` so it can't store `COV_FLAG_UNREACHABLE` (1 << 25). It works here because it happens to always be in a register but if the result were spilled to the stack or stored in a struct or global variable, a true value would mysteriously become false which is very hard to debug.

nickg (changes requested):

nickg (approved):

PR #622, opened 2023-02-16
Adding VHDL-2019 to OSVVM installation.

bpadalino:
Add a VHDL-2019 compiled installation to OSVVM when installing from nvc.


contrib/install-osvvm.tcl
```diff
@@ -2,3 +2,5 @@ source "Scripts/StartNVC.tcl"
 set ::osvvm::VhdlLibraryDirectory "${::env(NVC_INSTALL_DEST)}"
 set ::osvvm::VhdlLibrarySubdirectory "."
 source "OsvvmLibraries.pro"
+SetVHDLVersion 2019
```
nickg, line 5: Can we do this with a loop? E.g.

```tcl
foreach std {2008 2019} {
  ...
}
```

nickg (approved):

PR #624, opened 2023-02-17
Add parsing changes for LCS-2016-086.

bpadalino:
All interface lists can be ordered.

This allows for entities and subprograms to reference earlier elements in the interface list.

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -4643,8 +4643,9 @@ static void p_interface_constant_declaration(tree_t parent, tree_kind_t kind)
       add_interface(parent, d, kind);
       sem_check(d, nametab);
 
-      if (kind == T_GENERIC_DECL && standard() >= STD_08) {
+      if ((standard() >= STD_19) || (kind == T_GENERIC_DECL && standard() >= STD_08)) {
          // Generics are immediately visible in VHDL-2008
+         // Everything is immediately visible in VHDL-2019
          insert_name(nametab, d, NULL);
       }
    }
@@ -4691,6 +4692,10 @@ static void p_interface_signal_declaration(tree_t parent, tree_kind_t kind)
 
       add_interface(parent, d, kind);
       sem_check(d, nametab);
+
+      // 2019: Ports and generics are immediately visible
+      if (standard() >= STD_19)
+         insert_name(nametab, d, NULL);
    }
 }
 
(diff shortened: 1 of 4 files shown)


src/parse.c
```diff
@@ -4643,8 +4643,9 @@ static void p_interface_constant_declaration(tree_t parent, tree_kind_t kind)
       add_interface(parent, d, kind);
       sem_check(d, nametab);
 
-      if (kind == T_GENERIC_DECL && standard() >= STD_08) {
+      if ((standard() >= STD_19) || (kind == T_GENERIC_DECL && standard() >= STD_08)) {
          // Generics are immediately visible in VHDL-2008
+         // Everything is immediately visible in VHDL-2019
          insert_name(nametab, d, NULL);
```
nickg, line 4649: Although it's slightly redundant, I think it's clearer split over two cases:

```suggestion
      if (kind == T_GENERIC_DECL && standard() >= STD_08) {
         // Generics are immediately visible in VHDL-2008
         insert_name(nametab, d, NULL);
      }
      else if (standard() >= STD_19) {
         // Everything is immediately visible in VHDL-2019
         insert_name(nametab, d, NULL);
```

test/parse/vhdl2019.vhd
```diff
@@ -13,3 +13,10 @@ package pack is
     end record ;
 
 end package ;
+
+-- LCS-2016-086
+entity E is
+    generic (G1: INTEGER; G2: INTEGER := G1; G3, G4, G5, G6: INTEGER;G7:G6'SUBTYPE);
+    port (P1: STRING(G3 to G4); P2: STRING(P1'RANGE); P3: P1'SUBTYPE);
+    procedure X (Y1, Y2: INTEGER; Y3: INTEGER range Y1 to Y2; Y4: Y1'SUBTYPE);
+end E;
```
nickg, line 22: You need to extend `test_vhdl2019` to actually parse this (at the moment it only does the first entity). It should look like:

```c

START_TEST(test_vhdl2019)
{
   set_standard(STD_19);
   input_from_file(TESTDIR "/parse/vhdl2019.vhd");

   tree_t e1 = parse();
   fail_if(e1 == NULL);
   fail_unless(tree_kind(e1) == T_ENTITY);

   tree_t p1 = parse();
   fail_if(p1 == NULL);
   fail_unless(tree_kind(p1) == T_PACKAGE);

   tree_t e2 = parse();
   fail_if(e2 == NULL);
   fail_unless(tree_kind(e2) == T_ENTITY);

   fail_unless(parse() == NULL);

   fail_if_errors();
}
END_TEST
```

bpadalino: OK - this should be fixed in fa7a2b0.

I used the LCS change number so that modifications can be slightly easier to track as more features are developed for the test.

nickg (approved):

PR #625, opened 2023-02-18
Add LCS-2016-072b feature

bpadalino:
This allows a function to know the return type.

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -6148,7 +6148,8 @@ static tree_t p_subprogram_specification(void)
 
    // 2019:
    //  [ pure | impure ] function designator subprogram_header
-   //       [ [parameter] ( formal_parameter_list ) ] return [ return_identifier of ] type_mark
+   //       [ [parameter] ( formal_parameter_list ) ] 
+   //       return [ return_identifier of ] type_mark
 
    BEGIN("subprogram specification");
 
--- a/src/parse.c
+++ b/src/parse.c
@@ -4842,7 +4842,8 @@ static tree_t p_interface_function_specification(void)
 
    // 2019:
    // [ pure | impure ] function designator
-   //    [ [ parameter ] ( formal_parameter_list ) ] return [ return_identifier of ] type_mark
+   //    [ [ parameter ] ( formal_parameter_list ) ] 
+   //    return [ return_identifier of ] type_mark
 
    BEGIN("interface function specification");
 
--- a/src/parse.c
+++ b/src/parse.c
@@ -4873,7 +4873,7 @@ static tree_t p_interface_function_specification(void)
    if (peek_nth(2) != tOF)
       type_set_result(type, p_type_mark());
    else {
-      require_std(STD_19, "function knows return type") ;
+      require_std(STD_19, "function knows return type");
       ident_t id = p_identifier();
 
       consume(tOF);
@@ -6212,7 +6212,7 @@ static tree_t p_subprogram_specification(void)
       if (peek_nth(2) != tOF)
          type_set_result(type, p_type_mark());
       else {
-         require_std(STD_19, "function knows return type") ;
+         require_std(STD_19, "function knows return type");
 
          ident_t id = p_identifier();
 
--- a/src/parse.c
+++ b/src/parse.c
@@ -4874,20 +4874,24 @@ static tree_t p_interface_function_specification(void)
       type_set_result(type, p_type_mark());
    else {
       require_std(STD_19, "function knows return type") ;
-      ident_t id = p_identifier();
-
       tree_t t = tree_new(T_SUBTYPE_DECL);
       tree_set_loc(t, CURRENT_LOC);
+      ident_t id = p_identifier();
 
       consume(tOF);
 
       type_t tm = p_type_mark();
+
       tree_set_ident(t, id);
       tree_set_type(t, tm);
 
-      type_set_result(type, tm);
+      type_t sub = type_new(T_SUBTYPE);
+      type_set_ident(sub, id);
+      type_set_base(sub, tm);
+
       insert_name(nametab, t, id);
       sem_check(t, nametab);
+      type_set_result(type, sub);
    }
 
    tree_set_loc(d, CURRENT_LOC);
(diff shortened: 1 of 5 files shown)


src/parse.c
```diff
@@ -4865,7 +4869,29 @@ static tree_t p_interface_function_specification(void)
    }
 
    consume(tRETURN);
-   type_set_result(type, p_type_mark());
+   if (standard() < STD_19) {
+      if (peek_nth(2) == tOF)
```
nickg, line 4873: I'd switch this around and always parse the return identifier if `peek_nth(2) == tOF`. That way users who set the wrong standard version will get a single "requires VHDL-2019" error rather than a lot of unexpected token errors.

src/parse.c
```diff
@@ -4865,7 +4869,29 @@ static tree_t p_interface_function_specification(void)
...
+         require_std(STD_19, "function knows return type");
+      type_set_result(type, p_type_mark());
+   } else {
+      if (peek_nth(2) == tOF) {
+         ident_t id = p_identifier();
+
+         tree_t t = tree_new(T_SUBTYPE_DECL);
```
nickg, line 4880: Actually the declaration object should only go in the function body. For the specification we just need to create a new subtype `type_t` (I forgot to explain that bit). So something like:

```c
type_t sub = type_new(T_SUBTYPE);
type_set_id(sub, p_identifier());
type_set_base(sub, p_type_mark());

type_set_result(type, sub);
```

test/parse/vhdl2019.vhd
```diff
@@ -1,10 +1,28 @@
...
+--entity ent is
+--  port (
+--    a : in  boolean ;
+--    b : out boolean ;
+--  ) ;
+--end entity ;
+
+-- LCS-2016-072b: Function Knows Vector Size
+library ieee;
+use ieee.numeric_std.all;
```
nickg, line 11: Try to avoid the IEEE libraries in unit tests as they can be a little slow to load (the "check" library forks a new child process for each test, and there are hundreds of tests). In this case you could just make the return type `bit_vector` and then `return rv;` and it would still cover the same parser changes.

bpadalino: This should be covered with [9259d58](https://github.com/nickg/nvc/pull/625/commits/9259d584f2b8bd3ece23949ab3e8a33efc079379).

src/parse.c
bpadalino: This should be covered with 5c9b104.

src/parse.c
bpadalino: I don't quite understand what you mean by this.  We want to create a new `T_SUBTYPE` or `T_SUBTYPE_DECL`?

src/parse.c
```diff
@@ -6186,7 +6213,26 @@ static tree_t p_subprogram_specification(void)
...
+
+         consume(tOF);
+
+         type_t tm = p_type_mark();
+         tree_set_ident(t, id);
+         tree_set_type(t, tm);
+
+         type_set_result(type, tm);
+         insert_name(nametab, t, id);
+         sem_check(t, nametab);
```
nickg, line 6234:
```suggestion
         ident_t id = p_identifier();
         
         consume(tOF);
         
         type_t sub = type_new(T_SUBTYPE);
         type_set_id(sub, id);
         type_set_base(sub, p_type_mark());
         
         type_set_result(type, sub);
```

src/parse.c
nickg: > I don't quite understand what you mean by this. We want to create a new `T_SUBTYPE` or `T_SUBTYPE_DECL`?

Both. The `type_t` needs to get created in the specification and used as the result type, the `tree_t` needs to get created in the body and added to the declaration list.

src/parse.c
```diff
@@ -6218,14 +6222,23 @@ static tree_t p_subprogram_specification(void)
          type_set_result(type, p_type_mark());
       else {
          require_std(STD_19, "function knows return type") ;
+         tree_t t = tree_new(T_SUBTYPE_DECL);
```
nickg, line 6225: The previous version was actually correct: we don't want to create the `T_SUBTYPE_DECL` here but in `p_subprogram_body`. If you revert it to just creating the subtype `type_t` here, I can push something that does the second part - that's probably easier than trying to explain it in English :-).

nickg:
I just pushed something to create the subtype declaration in the right place. The subtype doesn't have any constraints at the moment so we get an error when declaring the `rv` variable. I need to think about how to represent that properly. So for now I suggest just adding that as an expected error and commit the parsing changes only.

nickg:
Can you try rebasing on the latest master? I want to see if the resolves the error with the formatting check.

bpadalino:
OK - I cherry-picked since I rebased against `master`.

bpadalino:
I updated the parsing tests, but I don't quite understand how to check for the errors just yet.  I also don't know how to add the parsed package such that the package body knows there's something there.

nickg:
> I updated the parsing tests, but I don't quite understand how to check for the errors just yet. 

Search for `expect_errors` in that file for examples. Something like:

```c
   const error_t expect[] = {
      { NNN, "error message" },
      { -1, NULL }
   };
   expect_errors(expect);
```

and change `fail_if_errors()` to `check_expected_errors()`.

> I also don't know how to add the parsed package such that the package body knows there's something there.

After parsing the package declaration in the unit test do:

```c
lib_put(lib_work(), package);
```

bpadalino:
Added the error check and rebased against current master.

nickg:
Is this one ready now or still a draft?

bpadalino:
This should be good to go.  Let me mark it ready for review.

bpadalino:
Sorry - hold on.  Let me change the feature to partial since it's parsing only.

src/parse.c
```diff
@@ -4840,6 +4840,10 @@ static tree_t p_interface_function_specification(void)
    // [ pure | impure ] function designator
    //    [ [ parameter ] ( formal_parameter_list ) ] return type_mark
 
+   // 2019:
+   // [ pure | impure ] function designator
+   //    [ [ parameter ] ( formal_parameter_list ) ] return [ return_identifier of ] type_mark
```
nickg, line 4845: Can you wrap the lines at 80 characters? (Here and below.)

```suggestion
   // [ pure | impure ] function designator
   //    [ [ parameter ] ( formal_parameter_list ) ] 
   //    return [ return_identifier of ] type_mark
```

src/parse.c
```diff
@@ -6127,6 +6146,10 @@ static tree_t p_subprogram_specification(void)
    //   | [ pure | impure ] function designator subprogram_header
    //       [ [parameter] ( formal_parameter_list ) ] return type_mark
 
+   // 2019:
+   //  [ pure | impure ] function designator subprogram_header
+   //       [ [parameter] ( formal_parameter_list ) ] return [ return_identifier of ] type_mark
```
nickg, line 6151: And this one :-)

```suggestion
   //  [ pure | impure ] function designator subprogram_header
   //       [ [parameter] ( formal_parameter_list ) ] 
   //       return [ return_identifier of ] type_mark
```

nickg (approved):

PR #629, opened 2023-02-21
Update entity name list parsing.

bpadalino:
When working with the [all08.vhdl](https://github.com/ghdl/ghdl/blob/c33f41e4612347d1fcbb9a37bc82249d7b8366eb/testsuite/sanity/004all08/all08.vhdl) file from the GHDL repository, there was a parsing error with the following:

```vhdl
  attribute user_attr of clear [std_logic_vector]: procedure is true;
```

The LRM states that the `entity_name_list` can be a list of `entity_designator` which could also include a `signature`.  It can also be the keywords `all` or `others`, if they are the last attribute for that particular class in the declarative part.  It's an error if they aren't the last.

With that being said, this gets past the parsing issue because of the other parsing examples in the file, but obviously things aren't correct with the tree entries.

`nvc` subsequently dies later in the analysis, but I believe it's more due to generics being applied than this parsing.

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -5264,13 +5264,51 @@ static class_t p_entity_class(void)
    }
 }
 
+static ident_t p_entity_designator()
+{
+   // entity_tag [ signature ]
+   ident_t id = p_identifier();
+
+   if (peek() == tLSQUARE) {
+      type_t constraints = p_signature();
+   }
+   return id;
+}
+
+static ident_list_t *p_entity_name_list(void)
+{
+   // entity_designator { , entity_designator } | others | all
+   ident_list_t *result = NULL;
+
+   switch (peek()) {
+   // LRM-2019 7.2 - Attribute Specification
+   case tOTHERS:
+      // Attribute applies to the names entities of the class declared in the
+      // immediately enclosing declarative part, provided that each such entity
+      // is not explicitly named in the entity name list of the previous
+      // attribute specification for the given attribute. Error if this isn't
+      // the last attribute for the given entity class.
+   case tALL:
+      // Attribute applies to all named entities of the class declared in the
+      // immediately enclosing declarative part. Error if this isn't the last
+      // attribute for the given entity class.
+      break;
+   default:
+      ident_list_push(&result, p_entity_designator(), last_loc);
+
+      while (optional(tCOMMA))
+         ident_list_push(&result, p_entity_designator(), last_loc);
+   }
+   return result;
+}
+
 static ident_list_t *p_entity_specification(class_t *class)
 {
    // entity_name_list : entity_class
 
    BEGIN("entity specification");
 
-   ident_list_t *ids = p_identifier_list();
+   ident_list_t *ids = p_entity_name_list();
 
    consume(tCOLON);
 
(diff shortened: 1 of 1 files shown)


src/parse.c
```diff
@@ -5245,13 +5245,51 @@ static class_t p_entity_class(void)
...
+   ident_list_t *result = NULL;
+
+   switch (peek()) {
+   // LRM-2019 7.2 - Attribute Specification
+   case tOTHERS:
+      // Attribute applies to the names entities of the class declared in the
+      // immediately enclosing declarative part, provided that each such entity
+      // is not explicitly named in the entity name list of the previous
+      // attribute specification for the given attribute. Error if this isn't
+      // the last attribute for the given entity class.
```
nickg, line 5271: I don't think there's any need to have all this text here, and these rules wouldn't be checked by the parser at any rate.

src/parse.c
```diff
@@ -5245,13 +5245,51 @@ static class_t p_entity_class(void)
...
+   case tALL:
+      // Attribute applies to all named entities of the class declared in the
+      // immediately enclosing declarative part. Error if this isn't the last
+      // attribute for the given entity class.
+      break;
```
nickg, line 5276: Probably needs a call to `consume(tALL)` here and similarly for `tOTHERS` above, otherwise you'll immediately get an unexpected token error for the next token?

src/parse.c
```diff
@@ -5266,6 +5304,10 @@ static void p_attribute_specification(tree_t parent, add_func_t addf)
    BEGIN("attribute specification");
 
    consume(tATTRIBUTE);
+   // entity_name_list ::=
+   //   entity_designator { , entity_designator }
+   //  |others
+   //  |all
```
nickg, line 5310: Doesn't belong here?

nickg (commented):
I think this is ok to go in with the modifications above (and a test). Those attributes won't work correctly but at least the situation will be slightly better than now.

src/parse.c
bpadalino: Agreed.  I'll do that, but what I was hoping to figure out here is how to convey the information that the target is `others` or `all` beyond the scope of this parsing function.

nickg: For now it can probably just use the hack that `p_instantiation_list` does and return NULL for `others` and the magic identifier `well_known(W_ALL)`. Longer term I want to rewrite most of the attribute handling code as there are multiple issues with it beyond just this one. I suspect the right thing to do is add a `T_DESIGNATOR` AST node to hold it but I'm not sure exactly how it should work.

src/parse.c
```diff
@@ -5264,13 +5264,47 @@ static class_t p_entity_class(void)
...
 }
 
+static ident_t p_entity_designator()
+{
+   // entity_tag [ signature ]
+   ident_t id = p_identifier();
+
+   if (peek() == tLSQUARE) {
+      // XXX: Review what to do here
+      type_t constraints = p_signature();
```
bpadalino, line 5274: I don't know what to do here with the signature.  Any suggestions?

nickg: I would just ignoring it for now with a warning. E.g. 

```suggestion
   if (peek() == tLSQUARE) {
      // XXX: Review what to do here
      (void)p_signature();
      warn_at(CURRENT_LOC, "sorry, signature in attribute entity name list is not yet supported");
```

It needs some more refactoring to handle properly.

nickg (approved):

PR #630, opened 2023-02-21
Add LCS-2016-055a (Optional Component Ending).

bpadalino:
Makes the `component` keyword at the end of a component declaration optional.

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -4842,7 +4842,7 @@ static tree_t p_interface_function_specification(void)
 
    // 2019:
    // [ pure | impure ] function designator
-   //    [ [ parameter ] ( formal_parameter_list ) ] 
+   //    [ [ parameter ] ( formal_parameter_list ) ]
    //    return [ return_identifier of ] type_mark
 
    BEGIN("interface function specification");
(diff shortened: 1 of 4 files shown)


src/parse.c
```diff
@@ -7501,7 +7505,11 @@ static tree_t p_component_declaration(void)
    pop_scope(nametab);
 
    consume(tEND);
-   consume(tCOMPONENT);
+   if (peek() != tCOMPONENT) {
+      require_std(STD_19, "optional end component");
+      optional(tCOMPONENT);
+   } else
```
nickg, line 7511: `else` goes on a new line. Can you try rebasing to see if the clang-format check is fixed now?

bpadalino: Rebased and pushed.

src/parse.c
```diff
@@ -7501,7 +7505,12 @@ static tree_t p_component_declaration(void)
    pop_scope(nametab);
 
    consume(tEND);
-   consume(tCOMPONENT);
+   if (peek() != tCOMPONENT) {
+      require_std(STD_19, "optional end component");
+      optional(tCOMPONENT);
+   }
```
nickg, line 7511: The `optional(tCOMPONENT)` is redundant: we already know at this point the next token is not `tCOMPONENT`.

```suggestion
   if (peek() != tCOMPONENT)
      require_std(STD_19, "optional end component");
```

nickg (approved):

PR #639, opened 2023-03-01
src: Add coverage-spec

Blebowski:
This MR adds `coverage-spec` argument to `elaborate`. A coverage specification file shall be passed
to this option. Coverage specification file allows controlling hierarchy / entity names on which coverage
is collected. Following is allowed within coverage specification file:

`(+|-)hierarchy <HIERARCHY>`  - Enables/disables collecting coverage for given hierarchy
`(+|-)block <BLOCK_NAME>`  - Enables/disables collecting coverage for given entity names / block names

Tests and manual edit still to be added in this MR.

### DIFF
--- a/src/lower.c
+++ b/src/lower.c
@@ -10817,17 +10817,23 @@ lower_unit_t *lower_instance(lower_unit_t *parent, cover_tagging_t *cover,
    lower_unit_t *lu = lower_unit_new(parent, vu, cover, block);
 
    if (cover_enabled(lu->cover, COVER_MASK_ALL)) {
-      lower_unit_t *it;
-      for (it = lu; it != NULL; it = it->parent) {
-         if (it->hier == NULL)
-            continue;
 
-         tree_t unit = tree_ref(it->hier);
-         if (tree_kind(unit) == T_ARCH) {
-            ident_t ename = ident_rfrom(tree_ident(tree_primary(unit)), '.');
-            cover_set_block_name(lu->cover, ename);
+      // Need to find corresponding T_HIER in advance before hier
+      // of lower_scope_t is assigned.
+      tree_t inst_hier = NULL;
+      for (int i = 0; i < tree_decls(block); i++) {
+         tree_t decl = tree_decl(block, i);
+         if (tree_kind(decl) == T_HIER) {
+            inst_hier = decl;
             break;
          }
+      }
+      assert(inst_hier != NULL);
+
+      tree_t unit = tree_ref(inst_hier);
+      if (tree_kind(unit) == T_ARCH) {
+         ident_t ename = ident_rfrom(tree_ident(tree_primary(unit)), '.');
+         cover_set_block_name(lu->cover, ename);
       }
 
       cover_add_tag(block, NULL, lu->cover, TAG_HIER, COV_FLAG_HIER_DOWN);
--- a/src/lower.c
+++ b/src/lower.c
@@ -10820,7 +10820,7 @@ lower_unit_t *lower_instance(lower_unit_t *parent, cover_tagging_t *cover,
       tree_t hier = tree_decl(block, 0);
       assert(tree_kind(hier) == T_HIER);
 
-      tree_t unit = tree_ref(inst_hier);
+      tree_t unit = tree_ref(hier);
       if (tree_kind(unit) == T_ARCH) {
          ident_t ename = ident_rfrom(tree_ident(tree_primary(unit)), '.');
          cover_set_block_name(lu->cover, ename);
--- a/src/lower.c
+++ b/src/lower.c
@@ -10817,18 +10817,8 @@ lower_unit_t *lower_instance(lower_unit_t *parent, cover_tagging_t *cover,
    lower_unit_t *lu = lower_unit_new(parent, vu, cover, block);
 
    if (cover_enabled(lu->cover, COVER_MASK_ALL)) {
-
-      // Need to find corresponding T_HIER in advance before hier
-      // of lower_scope_t is assigned.
-      tree_t inst_hier = NULL;
-      for (int i = 0; i < tree_decls(block); i++) {
-         tree_t decl = tree_decl(block, i);
-         if (tree_kind(decl) == T_HIER) {
-            inst_hier = decl;
-            break;
-         }
-      }
-      assert(inst_hier != NULL);
+      tree_t hier = tree_decl(block, 0);
+      assert(tree_kind(hier) == T_HIER);
 
       tree_t unit = tree_ref(inst_hier);
       if (tree_kind(unit) == T_ARCH) {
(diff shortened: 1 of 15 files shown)


src/lower.c
```diff
@@ -10534,15 +10534,27 @@ static vcode_unit_t lower_concurrent_block(tree_t block, vcode_unit_t context)
...
+   lower_push_scope(block);
+
    if (cover_enabled(cover_tags, COVER_MASK_ALL)) {
       if (!context)
          cover_reset_scope(cover_tags, prefix);
-
       cover_push_scope(cover_tags, block);
+
+      lower_scope_t *it;
+      for (it = top_scope; it != NULL; it = it->down) {
```
nickg, line 10545: This needs rebasing as I refactored a lot of this code (see `lower_instance`).

Blebowski: OK, done. I have iterated to through the `decls` t find out the `T_HIER` that could lead me to the
entity name of the current instance. I needed to do this before `lower_decls` on the given instance,
since I wanted to to take into account toggle tags on ports (If I disable coverage collecting on an
entity, I don't want to collect coverage on ports of its instance).

I hope its not too much of duplication against `lower_decls` -> `lower_decl` -> `lower_inst`.

Blebowski:
MR is now ready for review.

src/lower.c
```diff
@@ -10805,8 +10805,28 @@ lower_unit_t *lower_instance(lower_unit_t *parent, cover_tagging_t *cover,
...
+      // of lower_scope_t is assigned.
+      tree_t inst_hier = NULL;
+      for (int i = 0; i < tree_decls(block); i++) {
+         tree_t decl = tree_decl(block, i);
+         if (tree_kind(decl) == T_HIER) {
+            inst_hier = decl;
+            break;
+         }
+      }
+      assert(inst_hier != NULL);
```
nickg, line 10820: It's always the first declaration so you can just do:

```suggestion
      tree_t hier = tree_decl(block, 0);
      assert(tree_kind(hier) == T_HIER);
```

src/rt/cover.c
```diff
@@ -52,12 +53,14 @@ typedef struct {
 
 typedef A(cover_tag_t) tag_array_t;
 typedef A(line_range_t) range_array_t;
+typedef A(text_buf_t*) tb_array_t;
```
nickg, line 56: This can just be an array of `char *` since you don't need to modify the strings after they are created.

src/rt/cover.c
```diff
@@ -458,6 +474,55 @@ void cover_reset_scope(cover_tagging_t *tagging, ident_t hier)
    tagging->hier = hier;
 }
 
+bool cover_should_emit_scope(cover_tagging_t *tagging, tree_t t)
```
nickg, line 477:
```suggestion
static bool cover_should_emit_scope(cover_tagging_t *tagging, tree_t t)
```

src/rt/cover.c
```diff
@@ -505,17 +570,25 @@ void cover_push_scope(cover_tagging_t *tagging, tree_t t)
...
       s->sig_pos = tagging->top_scope->sig_pos;
 
    tagging->top_scope = s;
    tagging->hier = ident_prefix(tagging->hier, name, '.');
 
+   s->emit = true;
+   if (tagging->spec) {
+      s->emit = false;
+      s->emit = cover_should_emit_scope(tagging, t);
+   }
```
nickg, line 586:
```suggestion
   s->emit = (tagging->spec && cover_should_emit_scope(tagging, t));
```

src/rt/cover.c
```diff
@@ -715,6 +794,77 @@ void cover_count_tags(cover_tagging_t *tagging, int32_t *n_stmts,
...
+{
+   cover_exclude_ctx_t ctx = {};
+
+   ctx.ef = fopen(path, "r");
+   if (ctx.ef == NULL)
+      fatal_errno("failed to open coverage specification file: %s\n", path);
+
+   tagging->spec = xcalloc(sizeof(cover_spec_t));
+
+   char *delim = " ";
```
nickg, line 807:
```suggestion
   const char *delim = " ";
```

src/rt/cover.c
```diff
@@ -715,6 +794,77 @@ void cover_count_tags(cover_tagging_t *tagging, int32_t *n_stmts,
...
+
+      // Parse line as space separated tokens
+      for (char *tok = strtok(ctx.line, delim); tok; tok = strtok(NULL, delim)) {
+         bool include;
+         if (tok[0] == '+')
+            include = true;
+         else if (tok[0] == '-')
+            include = false;
+         else
+            fatal_at(&ctx.loc, "Coverage specification command must "
```
nickg, line 830:
```suggestion
            fatal_at(&ctx.loc, "coverage specification command must "
```

Start error messages with a lower case character.

src/rt/cover.c
```diff
@@ -715,6 +794,77 @@ void cover_count_tags(cover_tagging_t *tagging, int32_t *n_stmts,
...
+            fatal_at(&ctx.loc, "invalid command: $bold$%s$$", tok);
+
+         char *hier = strtok(NULL, delim);
+         if (!hier)
+            fatal_at(&ctx.loc, "%s name missing", tok);
+
+         text_buf_t *tb = tb_new();
+         tb_printf(tb, "%s", hier);
+         tb_upcase(tb);
+         APUSH(*arr, tb);
```
nickg, line 857:
```suggestion
         APUSH(*arr, xstrdup(hier));
```

See the comment above about using `char *`.

src/rt/cover.c
Blebowski: Is there actually some performance / code size benefit? Or is it just for clarity?

src/rt/cover.c
Blebowski: srry

src/lower.c
```diff
@@ -10805,8 +10805,18 @@ lower_unit_t *lower_instance(lower_unit_t *parent, cover_tagging_t *cover,
 
    lower_unit_t *lu = lower_unit_new(parent, vu, cover, block);
 
-   if (cover_enabled(lu->cover, COVER_MASK_ALL))
+   if (cover_enabled(lu->cover, COVER_MASK_ALL)) {
+      tree_t hier = tree_decl(block, 0);
+      assert(tree_kind(hier) == T_HIER);
+
+      tree_t unit = tree_ref(inst_hier);
```
nickg, line 10812:
```suggestion
      tree_t unit = tree_ref(hier);
```

src/rt/cover.c
nickg: It's a pointer to a read-only string: if the program writes through this it will segfault. Better to get a compile error instead.

nickg (approved):

PR #647, opened 2023-03-11
Cond analysis directive

Blebowski:
This MR adds `--define` analysis option that supplies user defined compiler directive.
Compiler directives are held in an array, since amount of directives is likely small.
If this is not true, it can be later changed to hash table.

Duplicit directives are treated as errors to avoid ambiguity. However, firing error on
duplicit directives prevents "overloading" of directives when analyse command is
assembled from some "layered" scripting system (e.g. in VUnit per testbench define
can't be overriden by "per-test" define). Maybe a warning would be more approriate?

Note: Manual update still TBD -> MR is draft.

### DIFF
--- a/nvc.1
+++ b/nvc.1
@@ -203,7 +203,7 @@ relaxed by later standards.  See the
 .Sx RELAXED RULES
 section below for details.
 .It Fl -define Ns = Ns Ar name Ns = Ns Ar value
-Define a conditional analysis identifier (VHDL 2019). Option can be
+Define a conditional analysis identifier (VHDL-2019).  Option can be
 used multiple times to define multiple identifiers.
 .El
 .\" ------------------------------------------------------------
--- a/nvc.1
+++ b/nvc.1
@@ -202,7 +202,8 @@ Disable certain pedantic LRM conformance checks or rules that were
 relaxed by later standards.  See the
 .Sx RELAXED RULES
 section below for details.
-.It Fl -define Ns = Ns Ar name Ns = Ns Ar value
+.\" -D, --define
+.It Fl D Ar name Ns = Ns Ar value , Fl -define Ns = Ns Ar name Ns = Ns Ar value
 Define a conditional analysis identifier (VHDL-2019).  Option can be
 used multiple times to define multiple identifiers.
 .El
--- a/nvc.1
+++ b/nvc.1
@@ -202,6 +202,9 @@ Disable certain pedantic LRM conformance checks or rules that were
 relaxed by later standards.  See the
 .Sx RELAXED RULES
 section below for details.
+.It Fl -define Ns = Ns Ar name Ns = Ns Ar value
+Define a conditional analysis identifier (VHDL 2019). Option can be
+used multiple times to define multiple identifiers.
 .El
 .\" ------------------------------------------------------------
 .\" Elaboration options
--- a/src/nvc.c
+++ b/src/nvc.c
@@ -110,7 +110,7 @@ static void missing_argument(const char *what, char **argv)
    fatal("%s option $bold$%s$$ requires an argument", what, argv[optind - 1]);
 }
 
-static void parse_cond_analysis_identifier(char *optarg)
+static void parse_conda_id(char *optarg)
 {
    char *name = optarg;
    char *value = NULL;
@@ -181,7 +181,7 @@ static int analyse(int argc, char **argv)
          opt_set_int(OPT_RELAXED, 1);
          break;
       case 'd':
-         parse_cond_analysis_identifier(optarg);
+         parse_conda_id(optarg);
          break;
       default:
          abort();
--- a/src/nvc.c
+++ b/src/nvc.c
@@ -110,22 +110,14 @@ static void missing_argument(const char *what, char **argv)
    fatal("%s option $bold$%s$$ requires an argument", what, argv[optind - 1]);
 }
 
-static void parse_conda_id(char *optarg)
+static void parse_pp_define(char *optarg)
 {
-   char *name = optarg;
-   char *value = NULL;
+   char *eq = strchr(optarg, '=');
+   if (eq == NULL)
+      fatal("$bold$--define$$ argument must be KEY=VALUE");
 
-   while (*optarg) {
-      if (*optarg == '=') {
-         *optarg = '\0';
-         optarg++;
-         value = optarg;
-      }
-      else
-         optarg++;
-   }
-
-   conda_id_add(name, value);
+   *eq = '\0';
+   pp_defines_add(optarg, eq + 1);
 }
 
 static int analyse(int argc, char **argv)
@@ -146,7 +138,7 @@ static int analyse(int argc, char **argv)
    int c, index = 0;
    const char *spec = ":";
 
-   conda_id_init();
+   pp_defines_init();
 
    while ((c = getopt_long(next_cmd, argv, spec, long_options, &index)) != -1) {
       switch (c) {
@@ -181,7 +173,7 @@ static int analyse(int argc, char **argv)
          opt_set_int(OPT_RELAXED, 1);
          break;
       case 'd':
-         parse_conda_id(optarg);
+         parse_pp_define(optarg);
          break;
       default:
          abort();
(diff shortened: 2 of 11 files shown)


nickg:
> Compiler directives are held in an array, since amount of directives is likely small.
If this is not true, it can be later changed to hash table.

I would just use a hash table (`shash_t`). It's likely to be faster even for relatively small numbers of directives.

Blebowski:
OK, I will adjust the implementation. Btw. with regards to the directives, it is not possible to specify check
the directive was defined, but not checking for its value, right? (like in Verilog or C)

e.g.:

```
`if USER_DEFINED_DIRECTIVE
          <code_in_between>
`end
```

nickg:
I think that's a limitation of VHDL-2019: it doesn't include an `` `ifdef`` or similar operator.

nvc.1
```diff
@@ -202,6 +202,9 @@ Disable certain pedantic LRM conformance checks or rules that were
 relaxed by later standards.  See the
 .Sx RELAXED RULES
 section below for details.
+.It Fl -define Ns = Ns Ar name Ns = Ns Ar value
+Define a conditional analysis identifier (VHDL 2019). Option can be
```
nickg, line 206:
```suggestion
Define a conditional analysis identifier (VHDL-2019).  Option can be
```

Put two spaces between sentences and I normally hyphenate VHDL-XXXX.

src/nvc.c
```diff
@@ -110,6 +110,24 @@ static void missing_argument(const char *what, char **argv)
    fatal("%s option $bold$%s$$ requires an argument", what, argv[optind - 1]);
 }
 
+static void parse_cond_analysis_identifier(char *optarg)
+{
+   char *name = optarg;
```
nickg, line 115: This does something a bit odd when the option has multiple `=`s. I think it would be better to just use `strchr` like this:

```c
char *eq = strchr(optarg, '=');
if (eq == NULL)
   fatal("$bold$--define$$ argument must be KEY=VALUE");

*eq = '\0';
conda_id_add(optarg, eq + 1);
```

src/nvc.c
```diff
@@ -120,13 +138,16 @@ static int analyse(int argc, char **argv)
       { "psl",             no_argument,       0, 'P' },
       { "relax",           required_argument, 0, 'X' },
       { "relaxed",         no_argument,       0, 'R' },
+      { "define",          required_argument, 0, 'd' },
```
nickg, line 141:
```suggestion
      { "define",          required_argument, 0, 'D' },
```

src/nvc.c
```diff
@@ -120,13 +138,16 @@ static int analyse(int argc, char **argv)
...
       { 0, 0, 0, 0 }
    };
 
    const int next_cmd = scan_cmd(2, argc, argv);
    int c, index = 0;
    const char *spec = ":";
```
nickg, line 147:
```suggestion
   const char *spec = ":D:";
```

Let's also support `DKEY=VALUE`.

src/scan.c
```diff
@@ -38,16 +39,17 @@ typedef struct {
...
-static const char     *file_start;
-static size_t          file_sz;
-static const char     *read_ptr;
-static hdl_kind_t      src_kind;
-static loc_file_ref_t  file_ref = FILE_INVALID;
-static int             colno;
-static int             lineno;
-static int             lookahead;
-static int             pperrors;
-static cond_stack_t    cond_stack;
```
nickg, line 50: Might as well leave the indentation as it was for the unchanged lines as it can still line up.

src/scan.c
```diff
@@ -210,22 +212,31 @@ const char *token_str(token_t tok)
...
+   shash_free(cond_idents);
+}
+
+void conda_id_add(const char *name, const char *value)
+{
+   char *existing_val = (char*) shash_get(cond_idents, name);
+   if (existing_val)
+      errorf("conditional analysis directive '%s' already defined (%s)",
+             name, existing_val);
+   shash_put(cond_idents, name, (char*)value);
```
nickg:
```suggestion
   shash_put(cond_idents, name, xstrdup(value));
```

Although it works for all the current usages, it's safer to make a copy of the string. Also don't need a cast as any pointer can be passed to `void *`.

src/scan.c
```diff
@@ -210,22 +212,31 @@ const char *token_str(token_t tok)
...
+   conda_id_add("TOOL_NAME",    PACKAGE_NAME);
+   conda_id_add("TOOL_EDITION", "");
+   conda_id_add("TOOL_VERSION", PACKAGE_VERSION);
+
+}
+
+void conda_id_exit()
+{
+   shash_free(cond_idents);
+}
```
nickg: Don't need this function: usually it's not necessary to free static data that is only allocated once at start-up.

src/scan.h
```diff
@@ -45,6 +45,10 @@ hdl_kind_t source_kind(void);
 token_t processed_yylex(void);
 const char *token_str(token_t tok);
 
+void conda_id_init();
+void conda_id_exit();
+void conda_id_add(const char *ident, const char *value);
```
nickg, line 50: I preferred the name before: "conda" doesn't really mean anything. How about `preprocessor_define_*`? E.g. `preprocessor_define_add()`. 

Also `void foo();` means `foo` takes an unknown number of arguments (i.e. `foo(1, 2, 3)` is allowed). Use `void foo(void);` instead if it takes no arguments (this is different to C++).

test/test_parse.c
```diff
@@ -3247,6 +3247,7 @@ END_TEST
 
 START_TEST(test_cond1)
 {
+   conda_id_init();
```
nickg, line 3250: Just do this once in `main()` in `test/unit_test.c`.

src/scan.c
```diff
@@ -210,22 +212,31 @@ const char *token_str(token_t tok)
...
-      return PACKAGE_VERSION;
-   else
-      return NULL;
+   cond_idents = shash_new(16);
+
+   conda_id_add("VHDL_VERSION", standard_text(standard()));
+   conda_id_add("TOOL_TYPE",    "SIMULATION");
+   conda_id_add("TOOL_VENDOR",  PACKAGE_URL);
+   conda_id_add("TOOL_NAME",    PACKAGE_NAME);
+   conda_id_add("TOOL_EDITION", "");
```
nickg, line 223: Can you put a non-empty string here? I noticed one of the VHDL-2019 compliance tests is checking for it. I've no idea what `TOOL_EDITION` is supposed to be - maybe you could check Aldec or similar.

nickg (changes requested):

src/scan.c
nickg: Riviera Pro has "2022.04 Linux64" so maybe just make this one PACKAGE_VERSION too.

src/scan.h
Blebowski: I wanted something shorter, the previous one was way too long. True that `conda` does not mean anything.
WBU `pp_defines_init` and `pp_defines_add` ?

src/nvc.c
Blebowski: Not really, I think the `val` from `options` struct should be unique, and `D` is there already for `--dump-llvm`.

src/nvc.c
Blebowski: See the comment above. Should I change the `dump-llvm` to something else than `D` ?

src/scan.h
nickg: Yes that's ok.

src/nvc.c
nickg: Actually that option can be deleted: we don't generate any LLVM at analysis time any more.

test/regress/cond4.sh
```diff
@@ -0,0 +1,8 @@
+set -xe
```
nickg, line 1: I would just leave it as it was before without the shell script and test a single define. There's no reason to test two unless there's likely to be a bug there (I don't see why there would be).

Blebowski: Well, I always try to cover every case, thats why I changed it. I can revert it, if you want, but I would preffer testing both, despite chance of bug is small here.

nickg: Let's just test one define without the shell script. Then the test can run on Windows and there is less overhead.

PR #656, opened 2023-03-20
src: Draft of parsing of PSL property and sequence declarations.

Blebowski:


Blebowski:
Adds parsing of `sequence` and `property` declaration. So far draft only, needs to be revisited and tests added. Published here to get some  comments on:
- Tokens `const` and `boolean`. Does it really make sense as tokens? Shouldn't this be done as set of some predefined types for PSL declarations?
- Semantics of `boolean`. When actually instantiating the sequence/property, is it legal to pass e.g. bit or `std_logic` so to argument of type `boolean` of property or sequence?
- From parsing of `T_FUNC_DECL` it seemed to me that "parameters in declaration" are added as `T_PORTS`, while actual parameters passed to the function call / module instance are linked with `T_PARAMS`. Is this correct ?
- Conserning the Identifier, Is it prefferable to set the ident for the `T_PSL` tree_t , or for the underlying `T_PROPERTY_DECL` psl_node_t ?

tmeissner:
>  Semantics of `boolean`. When actually instantiating the sequence/property, is it legal to pass e.g. bit or `std_logic` so to argument of type `boolean` of property or sequence?

Yes, I think so. The PSL 2010 IEE std. defines:

**6.5.1.1 PSL formal parameter type classes**
In the VHDL flavor, both bit and std_ulogic map to Boolean type class, but it is an error to pass expressions of type bit
and std_ulogic to formal parameters b0 and b1 respectively in this example if the '=' operator is not defined for operands
of type bit and std_ulogic.

nickg:
>     * Tokens `const` and `boolean`. Does it really make sense as tokens? Shouldn't this be done as set of some predefined types for PSL declarations?

I think it makes sense to do it this way for now.

> 
>     * From parsing of `T_FUNC_DECL` it seemed to me that "parameters in declaration" are added as `T_PORTS`, while actual parameters passed to the function call / module instance are linked with `T_PARAMS`. Is this correct ?

Yes.

> 
>     * Conserning the Identifier, Is it prefferable to set the ident for the `T_PSL` tree_t , or for the underlying `T_PROPERTY_DECL` psl_node_t ?

It should go in the `psl_node_t`.

src/parse.c
```diff
@@ -10480,15 +10533,118 @@ static tree_t p_psl_clock_declaration(void)
...
+   BEGIN("PSL property declaration");
+
+   scan_as_psl();
+
+   psl_node_t decl = psl_new(P_PROPERTY_DECL);
+   consume(tPROPERTY);
+
+   scan_as_vhdl();
+   tree_set_ident(parent, p_identifier());
+   scan_as_psl();
```
nickg, line 10560:
```suggestion
   tree_set_ident(parent, p_identifier());
```

Switching the mode here shouldn't be necessary as `tID` will be returned in either.

src/parse.c
```diff
@@ -10480,15 +10533,118 @@ static tree_t p_psl_clock_declaration(void)
...
+   BEGIN("PSL sequence declaration");
+
+   scan_as_psl();
+
+   psl_node_t decl = psl_new(P_SEQUENCE_DECL);
+   consume(tSEQUENCE);
+
+   scan_as_vhdl();
+   tree_set_ident(parent, p_identifier());
+   scan_as_psl();
```
nickg, line 10597:
```suggestion
   tree_set_ident(parent, p_identifier());
```

src/parse.c
```diff
@@ -10480,15 +10533,118 @@ static tree_t p_psl_clock_declaration(void)
...
+
+   scan_as_vhdl();
+
+   return decl;
+}
+
+static tree_t p_psl_declaration(void)
+{
+   // PSL_Declaration ::=
+   //      Property_Declaration
```
nickg, line 10622:
```suggestion
   // Property_Declaration
```

I normally leave off the first part of the grammar production in the comments.

src/tree.c
```diff
@@ -334,7 +334,7 @@ static const imask_t has_map[T_LAST_TREE_KIND] = {
    (I_IDENT | I_ASSOCS | I_STMTS | I_DECLS),
 
    // T_PSL
-   (I_IDENT | I_FOREIGN),
+   (I_IDENT | I_PORTS | I_FOREIGN),
```
nickg, line 337: This is not actually used, or necessary I think. The port list should go in the `psl_node_t`.

Blebowski:
```
Another thing: What's with parsing of remaining parameter types?

From PSL IEEE 2010 6.5.1.1 (HDL formal parameter types):

bit Bit expression (refer to 5.1.1)
bitvector BitVector expression (refer to 5.1.3)
numeric Numeric expression (refer to 5.1.4)
string String expression (refer to 5.1.5)
From PSL IEEE 2010 6.5.1.2 (PSL formal parameter type classes):

hdltype
```

Hi @tmeissner , I am using PSL 1.1 document from Accelera as reference for implementation.
I don't have the IEEE 2005 PDF available. I hope this one represents the IEEE 2005 the closest.
I would like to start with the 2005 PSL version first, since this one is referenced by VHDL 2008 LRM.
I think it makes sense to start with earlier version of the standard and extend only later.

In the 2005, `bit` `bitvector` `numeric` or `string` are not defined.

tmeissner:
@Blebowski IIRC there are some small changes from Accelera's PSL LRM 1.1 document to the IEEE PSL standard of 2005. Let me look after these changes when I have access to the IEEE 2005 standard tomorrow 🙂

Blebowski:
> @Blebowski IIRC there are some small changes from Accelera's PSL LRM 1.1 document to the IEEE PSL standard of 2005. Let me look after these changes when I have access to the IEEE 2005 standard tomorrow slightly_smiling_face

Thanks :) I think I will anyway need to buy the standard at some point, since its not freely available anywhere.

Blebowski:
> @Blebowski All good, please read my comment as suggestion, not as order to implement these things wink

No order taken, thanks for the feedback :) Its been some time since using PSL, since in my current job
we swapped to SystemVerilog (and SVA), so I am completely forgetting about these tiny details.

tmeissner:
@Blebowski We also don't have the 2005 version of the IEEE 1850 document. But I have the book "A practival introduction to PSL" which states that it is based on that revision of the standard. And it contains a similar table of parameter types:

From the section **8.2 Named properties and SEREs** of that book:

> The formal parameters of a named property or SERE can be one of the
following types:
> * boolean: A Boolean expression
> * bit: A single bit
> * bitvector: A vector composed of bits
> * numeric: Any expression interpretable as an integer in the underlying flavor. For example, in the Verilog flavor, any bitvector that does not contain an unknown bit is interpretable as a numeric.
> * string: A string
> * sequence: A PSL sequence. The term sequence is used in PSL to denote one of the following: a braced SERE, a clocked SERE, a repeated SERE, or an instantiation of a named SERE.
> * property: A PSL property

>In addition, data types of the underlying flavor can be used as formal parameter types, by prepending the keyword hdltype to the type.

I recommend this book when it comes to PSL as it is one of the few about that topic and it is well written (by some of the creators of the Sugar language, which was the base for the PSL standardization process). Maybe izt's a better deal than buying the standard, as it also contains the formal syntax & demantic as appendix.

Blebowski:
Hi @tmeissner , thanks. It seems that these are one of the changes between IEEE and 1.1. then.
I will update the parsing code then.

Blebowski:
Hi @nickg , I have updated the parsing according to 2010 PSL standard. However, I am wondering, how to handle ports and entries to symbol table. Currently, I create `P_PORT_DECL` as an PSL alternative to `T_PORT_DECL`. However, If I want to
put entry to symbol table, I need `tree_t` port declaration (unless modifying the API in "names.h"). If so, than I would welcome
small guidance on how to proceed with the Symbol table changes to take into account PSL declarations and their ports.

Or, should I create a `T_PORT_DECL` as `I_FOREIGN` of `P_PORT_DECL`, or `P_PORT_DECL` does not make sense at all (and instead the upper `P_PROPERTY_DECL` should have `I_PORTS` that will point to `T_PORT_DECL` ? 

I tried to separate the PSL port from VHDL port, since I see that is the way you go in Verilog stub, since it seems
it is good idea not to have the ASTs too tangled between each other.

nickg:
>  or P_PORT_DECL does not make sense at all (and instead the upper P_PROPERTY_DECL should have I_PORTS that will point to T_PORT_DECL ?

I think this is the best way to do it (although it's better to use T_PARAM_DECL). You can change `psl_node_t psl_port(...)` to `tree_t psl_port(...)` and you probably won't need `psl_hdl_type()` at all.

Blebowski:
Hi, now more-less ready for review. Types are still not handled on `property` nor `sequence` parameters, but I guess this can be a separate work later on.

src/parse.c
```diff
@@ -10467,10 +10469,145 @@ static tree_t p_psl_directive(void)
...
 }
 
-static tree_t p_psl_clock_declaration(void)
+static type_t p_psl_subtype_indication(void)
+{
+   BEGIN("PSL subtype indication");
+
+   // Handle ambiguity in "p_subtype_indication". Two consecutive IDs
+   // may indicate resolution function, or only "type_mark" followed by
+   // actual parameter name.
```
nickg, line 10478: Couldn't you just call `p_subtype_indication()` here? Because subtype indication is `[ name ] type_mark [ constraint ]`.

src/parse.c
```diff
@@ -10467,10 +10469,145 @@ static tree_t p_psl_directive(void)
...
+   //
+   //    HDL_Type ::=
+   //       hdltype HDL_VARIABLE_TYPE
+   //
+   //    PSL_Type_Class ::=
+   //       boolean | bit | bitvector | numeric | string
+
+   BEGIN("PSL Parameter specification");
+
+   *class = C_MUTABLE;
```
nickg, line 10506: Is mutable the same or similar to the `variable` class in VHDL?

src/parse.c
```diff
@@ -10467,10 +10469,145 @@ static tree_t p_psl_directive(void)
...
+         type_t t = p_psl_subtype_indication();
+         scan_as_psl();
+         return t;
+      case tBOOLEAN:
+         *psl_type = PSL_TYPE_BOOLEAN;
+         return std_type(NULL, STD_BOOLEAN);
+      case tBIT:
+         *psl_type = PSL_TYPE_BIT;
+         return std_type(NULL, STD_BIT);
+         break;
```
nickg, line 10550: Don't need the `break` after `return`.

src/parse.c
```diff
@@ -10467,10 +10469,145 @@ static tree_t p_psl_directive(void)
...
+      case tNUMERIC:
+         *psl_type = PSL_TYPE_NUMERIC;
+         return std_type(NULL, STD_INTEGER);
+         break;
+      case tSTRINGK:
+         *psl_type = PSL_TYPE_STRING;
+         return std_type(NULL, STD_STRING);
+         break;
+      default:
+         one_of(tHDLTYPE, tBOOLEAN, tBIT, tBITVECTOR, tNUMERIC, tSTRINGK);
```
nickg, line 10564: This is covered by the previous call to `one_of` so can just return NULL here or let it fall through to the return below.

Does returning NULL cause any problems later if the input is invalid? Normally I return a dummy type `T_NONE` in cases where there was an error in the type. E.g. `type_new(T_NONE)`. That prevents cascading errors (it compares equal to every other type) and avoid having to check `tree_has_type()` all the time.

src/parse.c
Blebowski: I think no. When you put:
`const hdltype integer my_fancy_parameter`

then `p_subtype_indication` will evaluate:
`if ((peek() == tID && peek_nth(2) == tID) || peek() == tLPAREN) `
as true, and try to parse `integer` as subtype. Instead, it should
parse `p_type_mark`.

I wanted to avoid changing `p_subtype_indication`.

src/parse.c
Blebowski: TBH, I am not sure.

Can a parameter to PSL property or sequence be VHDL `signal` and also `shared variable`?
If both are allowed, I would keep the `C_MUTABLE` since it can represent both.
If only `signal` is allowed, then I think using `signal` class would be intuitive.

Or, does `variable` class represent anything that is "changable", and not kind of VHDL parameter ?

tmeissner: I don't think that other object classes beside `signal` make sense as parameters. But I'm not sure if it is defined exactly.

Blebowski: I would expect that if it is not specified, passing shared variable could be supported, but I agree that it is rather exotic case (possibly with many corner-cases), so I changed it by default to `C_SIGNAL`.

src/parse.c
nickg: OK I see. It's not a problem in VHDL because the name comes before the type.

nickg (approved):

PR #673, opened 2023-04-15
PSL repeated SERE parsing.

Blebowski:


Blebowski:
Now ready for review.

This MR adds parsing of `Repeated_SERE` and `Clocked_SERE` from 2010 PSL LRM. It handles recursive grammar for `Sequence` -> `Repeated_SERE` -> `Sequence` with arbitrary depth of "repeat" operators as well as `Proc_Block`.

Parsing of `Sequence_Inst` executes `ident_t` look-ahead for next `tID` (new `peek_ident`) and checks if the follow-up
identifier refers to previously defined PSL sequence within given scope. If not found, it parses as if `Boolean`.

A new test `parse4` is coming next, still in this MR.

src/psl/psl-node.c
```diff
@@ -69,18 +69,23 @@ static const imask_t has_map[P_LAST_PSL_KIND] = {
...
+   // P_SEQUENCE_INST
+   (I_REF | I_PARAMS),
 };
 
 static const char *kind_text_map[P_LAST_PSL_KIND] = {
    "P_ASSERT", "P_ASSUME", "P_RESTRICT", "P_FAIRNESS", "P_COVER", "P_ALWAYS",
    "P_HDL_EXPR", "P_PROPERTY_DECL", "P_SEQUENCE_DECL", "P_CLOCK_DECL", "P_NEXT",
    "P_NEVER", "P_EVENTUALLY", "P_NEXT_A", "P_NEXT_E", "P_NEXT_EVENT",
-   "P_SERE", "P_IMPLICATION",
+   "P_SERE", "P_IMPLICATION", "P_PROPERTY_INSTANCE", "P_SEQUENCE_INST"
```
nickg, line 88:
```suggestion
   "P_SERE", "P_IMPLICATION", "P_PROPERTY_INST", "P_SEQUENCE_INST"
```

src/psl/psl-node.h
```diff
@@ -108,6 +115,17 @@ ident_t psl_ident(psl_node_t p);
 bool psl_has_ident(psl_node_t p);
 void psl_set_ident(psl_node_t p, ident_t i);
 
+unsigned psl_ranges(psl_node_t p);
+tree_t psl_range(psl_node_t p, unsigned n);
+void psl_add_range(psl_node_t p, tree_t r);
```
nickg, line 120: Let's instead make this `psl_repeat` as there is only ever zero or one rather than a list (add an `I_REPEAT` to to `src/object.h`).

I think we could also have a new `P_REPEAT` node that has a `psl_subkind` and a `psl_tree` for the count/range, and `P_SERE` has an optional `psl_repeat`.

src/lexer.l
```diff
@@ -509,6 +513,10 @@ STRING_K      ?i:string
 % /* These PSL tokens must be scanned in VHDL mode for look-ahead */
 "->"                     { TOKEN(tIFIMPL); }
 "<->"                    { TOKEN(tIFFIMPL); }
+"[*"                     { TOKEN(tTIMESRPT); }
+"[+]"                    { TOKEN(tPLUSRPT); }
+"[="                     { TOKEN(tGOTORPT); }
+"[->"                    { TOKEN(tARROWRPT); }
```
nickg, line 519: It seems weird that these are tokens rather than separate `[` and `->` but the PSL spec seems to define it this way so ok.

src/psl/psl-node.h
Blebowski: OK, I see, that makes sense. I was thinking about this "redundancy". I first wanted to use the array to store multiple consecutive repetitions, e.g. in case like: `a[*2][*4][*6]`, and keep a single `P_SERE` object. However, then I realized
that `Proc_Block` can arbitrarily alterate with the "repeat", so then there is no way to track what all should be repeated.

So, the `P_REPEAT` will have `value` that will point to:
- `P_SERE` in case of braced/clocked SERE.
- `P_HDL_EXPR` in case of boolean.

src/lexer.l
Blebowski: Yup, I had similar feeling. Due to similar reason I kept `[[` as two separate tokens (not mentioned in LRM as token), but no examples in LRM show space between these two. I guess it needs some experimenting with other simulators and mimicing the behavior.

src/psl/psl-node.h
nickg: I was actually thinking it would work the other way round, so that `P_SERE` would hold the `P_REPEAT`.

```
a[*2][*4][*6]  ==  {{{a}[*2]}[*4]}[*6]

P_SERE
  \-----repeat----P_REPEAT---subkind--PSL_CONSEC_REPEAT
  |                   |
  |                   \-----value----T_LITERAL(6)
  |
  \-----operand[0]----P_SERE
                         \-----repeat----P_REPEAT----subkind--PSL_CONSEC_REPEAT
                         |                   |
                         |                   \------value---T_LITERAL(4)
                         |
                         \-----operand[0]----P_SERE (etc.)
```

Blebowski: OK that makes sense, the first object encountered during tree traversal is then always `P_SERE`.

Blebowski: But still the `P_REPEAT` needs at least `left` and `right`, since repeat can be range (instead of plain `value`).

nickg: You can just use a `T_RANGE` tree object and assign it to the value.

src/psl/psl-node.c
```diff
@@ -69,18 +69,26 @@ static const imask_t has_map[P_LAST_PSL_KIND] = {
    (I_SUBKIND | I_VALUE),
 
    // P_SERE
-   (I_PARAMS | I_CLOCK),
+   (I_SUBKIND | I_PARAMS | I_CLOCK | I_REPEAT | I_DECLS),
```
nickg, line 72: The `I_SUBKIND` isn't used any more?

Blebowski: That is true at the moment. As the very next step, I want to add parsing of all SERE kinds (`Compound_SERE`, `SERE concatenation` and `SERE fusion`), so It will be there. I already have it locally, just not pushed yet. I can remove it for this MR if you want, but It will come soon again.

nickg (approved):
There's some merge conflicts with the commits I just pushed.

PR #679, opened 2023-04-25
src: Add parsing of Compound SERE

Blebowski:
Adds parsing of `SERE` and `Compound_SERE`. SEREs are parsed recursively. If a SERE
contains chain of SEREs connectd with the same operator (e.g. `{a;b;c;d;e}`) the top most
SERE is a single SERE.

### DIFF
--- a/src/psl/psl-fsm.c
+++ b/src/psl/psl-fsm.c
@@ -66,7 +66,7 @@ static void build_implication(psl_fsm_t *fsm, fsm_state_t *state, psl_node_t p)
    case P_NEXT:
       {
          fsm_state_t *new = add_state(fsm);
-         add_edge(state, new, EDGE_NEXT, psl_operand(psl_operand(p, 0), 0));
+         add_edge(state, new, EDGE_NEXT, psl_operand(p, 0));
          build_node(fsm, new, psl_value(rhs));
       }
       break;
--- a/src/psl/psl-fsm.c
+++ b/src/psl/psl-fsm.c
@@ -66,13 +66,35 @@ static void build_implication(psl_fsm_t *fsm, fsm_state_t *state, psl_node_t p)
    case P_NEXT:
       {
          fsm_state_t *new = add_state(fsm);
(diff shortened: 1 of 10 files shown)


src/psl/psl-fsm.c
```diff
@@ -76,6 +76,28 @@ static void build_implication(psl_fsm_t *fsm, fsm_state_t *state, psl_node_t p)
...
+   if (ops == 1)
+      return;
+
+   for (int i = 1; i < ops; i++) {
+      psl_node_t rhs = psl_operand(p, i);
+      switch (psl_subkind(p)) {
+      case PSL_SERE_CONCAT:
+         fsm_state_t *new = add_state(fsm);
+         add_edge(state, new, EDGE_NEXT, psl_value(rhs));
+         build_node(fsm, new, rhs);
```
nickg, line 94:
```suggestion
         {
            fsm_state_t *new = add_state(fsm);
            add_edge(state, new, EDGE_NEXT, psl_value(rhs));
            build_node(fsm, new, rhs);
         }
         break;
```

test/test_psl.c
```diff
@@ -229,11 +298,11 @@ START_TEST(test_dump)
    tb_rewind(tb);
 
    psl_dump(tree_psl(tree_stmt(a, 1)));
-   ck_assert_str_eq(tb_get(tb), "assert never B");
+   ck_assert_str_eq(tb_get(tb), "assert never {B}");
```
nickg, line 301: This doesn't look right: a Boolean in an FL property shouldn't be parsed as a SERE.

Blebowski:
I have adjusted the sequence parsing. It works like so now:
- Plain `Boolean` is parsed as `P_HDL_EXPR`.
- Sequence instance is parsed as `P_SEQUENCE_INST`, repeat scheme can be set also to sequence instance.
- If a boolean is extended with repeat scheme, it becomes `P_SERE`.
- Nested repeat schemes create a new `P_SERE`.

nickg (approved):
Looks OK, thanks.

PR #691, opened 2023-05-13
WIP: Add parsing of PSL "union" and "built-in-function" calls.

Blebowski:
This MR adds basics of parsing of builtin PSL functions as in PSL LRM 2010.
A new tokens are added for functions.


src/parse.c
```diff
@@ -465,6 +466,14 @@ static int _one_of(int dummy, ...)
    }
 }
 
+static bool peek_psl_built_in(void)
+{
+   if (scan(tNEXT, tPREV, tSTABLE, tROSE, tFELL, tENDED, tISUNKNOWN,
+            tCOUNTONES, tONEHOT, tONEHOT0, tNONDET, tNONDETV))
+      return true;
+   return false;
```
nickg, line 474:
```suggestion
   return scan(tNEXT, tPREV, tSTABLE, tROSE, tFELL, tENDED, tISUNKNOWN,
               tCOUNTONES, tONEHOT, tONEHOT0, tNONDET, tNONDETV);
```

src/psl/psl-node.h
```diff
@@ -42,6 +42,9 @@ typedef enum {
    P_REPEAT,
    P_PROPERTY_INST,
    P_SEQUENCE_INST,
+   P_UNION,
+   P_BUILT_IN_FUNC,
```
nickg, line 46:
```suggestion
   P_BUILTIN_FUNC,
```

I think it's more common to write "bultin" as a single word in technical usage.

Blebowski:
Hi,

I noticed the original approach to parse `Built_In_Psl_Function` will not work. The reason is, that if the builtin function call needs to be parsed as a terminal, since it can be part of expression (e.g. `{a='1';prev{b}='0'}`). Pretty much any builtin function returning non-boolean return value, is likely to be used in expression to obtain the boolean that is used to evaluate progress/finish of SERE or sequence.

Therefore, I ended up parsing the expression underlying `p_psl_condtion` in PSL mode. When treating built-in function calls as tokens, then it is easy to non-invasively handle this in `p_primary`. I am still trying to figure out types though.

Generally, I think the built-in functions could be resolved by many different approaches, e.g.:
1. Implement the function semantics in VHDL (in an internal, user invisible package) in some global scope. Make this function available only during scope that parses inside `T_PSL`. This would benefit from VHDL infrastructure of the compiler. Calls of PSL builtin function would be as if regular function calls (e.g. rising_edge).
2. Parse builtin functions as `T_ID`, and pre-form `well_known` identifiers, and create some sort of "proxy" functions that would only have `T_FCALL`, but no real `T_FUNC_DECL`. Create logic of the built-in functions during "lower" phase.
3. Parse builtin functions as separate tokens, and create dedicated objects in PSL AST with function type given as subkind. In the expression, hold `T_FCALL` and point to `P_BUILTIN` via `T_PSL` to get to the actuall parameter values.  
4. Completely avoid holding this in PSL ast, and only create `T_FCALL`, resolve type, and hold parameter values as regular `tree_t` objects. This has a trouble with parameters that are `sequence`, so there would need to be a bridge from PSL AST to VHDL AST.

In general, the options of implementing this, range from almost native VHDL implementation (1) up to a completely PSL AST implementation (as separate from VHDL AST as possible). I am curious which approach would you take?

tmeissner:
I think, 1) would also be useful approach for the built-in `ended(SERE)` function which can be used in regular VHDL code in all simulators I know.

Blebowski:
I still think this makes sense only for some PSL functions. There are problems with:
`prev` needs to hold FIFO of its input values sampled with specified clock expression. This FIFO needs to be persistent across function calls, but only from a single place. I dont know how to do that.

`next` is tricky too, since it needs to get next value of an expression, so it may lead to something like an entity with postponed process in it.

`ended` needs to input arbitrary PSL sequence, so that would lead somehow converting the sequence to an fsm inside the function, that is quiet crazy... Not even talking about how to convert arbitrary psl sequence to vhdl arguments that can be passed to vhdl function.

`nondet` and `nondet_vector` have value set arguments, so those would somehow need to be converted to vhdl arrays or something like that. Plus, there can be variable number of ranges specified, so the function would need to implement randomizer that picks the value from input arguments. I am not sure how to implement that.

others are less tricky to do. I need to rethink this.

nickg:
Are those functions valid in the simple subset? I'm also not sure how they would be possible to implement efficiently. I think they can be ignored for now.

tmeissner:
I don't think that next() and nondet*() are included in the simple subset. These functions are more useful for formal verification, I assume. I haven't used a tool yet which supports these 3 functions.

nickg:
For `ended()` I think it would be ok to implement if the argument is just a PSL endpoint name e.g. `ended(endpoint_name)` because `endpoint_name` would be parsed as an identifier and then looked up as normal.

Blebowski:
The newest LRM does not have an `endpoint` at all, it is only in the Accellera PSL standards. AFAIK, the `ended()` serves as a replacement of `endpoint`.

nickg:
Can built-in function calls really be parsed as part of arbitrary VHDL expressions? I looked in the grammar in the 2010 LRM and it only appears in `HDL_or_PSL_Expression` as an alternative to `PSL_Expression` and `HDL_Expression` so it seems once you are inside `HDL_Expression` you can't parse a `Built_In_Function_Call`. In "A Practical Introduction to PSL" it just says a built-in function can be used "wherever a Boolean expression is expected".

nickg:
For `onehot()`, etc. which have trivial VHDL implementations we could provide a package containing them. Perhaps one that's implicitly `use`d inside PSL statements.

Blebowski:
OK, so the package can implement `onehot`, `onehot0`, `countones` and `isunknown` and make them visible only during parsing of PSL or HDL expression. For others I will stick to parsing as a dedicated keyword and parsing as a dedicated PSL node.

Blebowski:
I got to this after long time, and now I have it more-less finished. The built-in functions are split into two categories:
1. Functions parsed as new `P_BUILTIN_FUNC`
2. Functions implemented in NVC support package for PSL.

I have added simple test for already implemented built-in functions.

I have also added parsing of parameterized SEREs (so far without a test).

lib/nvc/psl_support.vhd
```diff
@@ -0,0 +1,88 @@
...
+
+    function onehot(
+        val : in bit_vector
+    ) return boolean;
+
+    ---------------------------------------------------------------------------
+    -- Returns true if there is zero or one occurence of '1' or 'H' in the
+    -- input vector.
+    ---------------------------------------------------------------------------
+    function onehot0(
```
nickg, line 76: I think you could probably make one version that works for both -93 and -08 by using the conditional analysis feature (which works in all standard versions). See `lib/synopsys/std_logic_misc.vhd` for an example.

src/common.h
```diff
@@ -19,6 +19,7 @@
 #define _COMMON_H
 
 #include "tree.h"
+#include "psl/psl-node.h"
```
nickg, line 22: Not needed?

src/parse.c
```diff
@@ -10656,6 +10671,61 @@ static tree_t p_generate_statement(ident_t label)
    }
 }
 
+static psl_node_t p_psl_value_range(void)
+{
+   //   Value
+   // | finite_Range
+
+   BEGIN("PSL Value set");
```
nickg, line 10679: "PSL value range"?

lib/nvc/psl_support.vhd
Blebowski: OK, thanks. I reworked it. I had no idea that conditional defines work also in 93.

nickg (approved):

PR #694, opened 2023-05-17
More VHPI support

sean-anderson-seco:
This adds enough support for VHPI that cocotb will run all the way through without it or nvc having a fatal error.

There is still some bug in the callback lifetimes, since I see some occasional crashes when shutting down the simulator.

With this series applied, the main stopper is `mode vhpiDepositPropagate not supported in vhpi_put_value` since deposits are not supported. Fixing that is a bit beyond my depth, since I'm not really familiar with the internals of this simulator.


src/elab.c
```diff
@@ -2046,15 +2046,18 @@ tree_t elab(tree_t top, jit_t *jit, cover_tagging_t *cover)
    case T_ENTITY:
       {
          tree_t arch = elab_pick_arch(tree_loc(top), top, &ctx);
+         tree_set_primary(e, arch);
```
nickg, line 2049: I think you can do this instead using the `T_HIER` object which is added to each `T_BLOCK` in the elaborated design as that has a reference to the design unit it was instantiated from, see my comment below.

src/vhpi/vhpi-model.c
```diff
@@ -1454,8 +1758,36 @@ void vhpi_build_design_model(tree_t top, rt_model_t *m)
 
    model = m;
 
+   tool = new_object(sizeof(c_tool), vhpiToolK);
+
+   c_designUnit *unit;
+   tree_t s = tree_primary(top);
```
nickg, line 1764:
```suggestion
   tree_t hier = tree_decl(b0, 0);
   assert(tree_kind(hier) == T_HIER);

   tree_t s = tree_ref(hier);
```

That would then work with instances inside the hierarchy. Sorry it's not very obvious how to do that. Not sure if you were aware but you can set `NVC_ELAB_VERBOSE=1` to print the elaborated design and then the `T_HIER` nodes are printed as `-- Instance foo`.

src/vhpi/vhpi-model.c
```diff
@@ -793,6 +1036,29 @@ vhpiIntT vhpi_get(vhpiIntPropertyT property, vhpiHandleT handle)
...
+         rt_signal_t *signal = vhpi_get_signal(decl);
+         if (signal == NULL)
+            return 0;
+
+         return signal_width(signal);
+      }
+
+   case vhpiArgcP:
+      if (obj->kind != vhpiToolK)
+            vhpi_error(vhpiInternal, &(obj->loc), "vhpiArgcP is only supported for tool objects");
```
nickg, line 1059:
```suggestion
         vhpi_error(vhpiInternal, &(obj->loc), "vhpiArgcP is only supported for tool objects");
```

src/vhpi/vhpi-model.c
```diff
@@ -347,10 +436,20 @@ static void init_abstractRegion(c_abstractRegion *r, tree_t t)
    r->LineOffset = loc->line_delta;
 
    r->Name = r->CaseName = new_string(istr(tree_ident(t)));
+   char *full LOCAL = xasprintf(":%s", r->Name);
+   r->FullName = r->FullCaseName = new_string(full);
+   printf("name=%s fullname=%s\n", r->Name, r->FullName);
```
nickg, line 441: Left a debugging printf in?

nickg (commented):
Thanks for doing this. I'll have a look at the `vhpiDepositPropagate` thing at the weekend. Have you got a fork of cocotb that you're using for testing?

> There is still some bug in the callback lifetimes, since I see some occasional crashes when shutting down the simulator.

Maybe try configuring with `--enable-asan` and see if that triggers.

src/rt/model.c
```diff
@@ -810,6 +810,11 @@ const void *signal_last_value(rt_signal_t *s)
    return s->shared.data + s->shared.size;
 }
 
+uint32_t signal_width(rt_signal_t *s)
+{
+   return s->nexus.width;
```
nickg, line 815:
```suggestion
   return s->shared.size / s->nexus.size;
```

There's a difference if sub-elements of a composite signal have to be tracked independently. For example if you have two processes:

```vhdl
signal x : std_logic_vector(2 downto 0);
...
p1: x(2 downto 1) <= a;
p2: x(0) <= b;
```

Then the `rt_signal_t` for `x` has two `rt_nexus_t` the first with width 2 and the second with width 1 but the overall signal width should be 3. (`shared.size` is in bytes rather than elements because it's more convenient for the generated code.)

sean-anderson-seco:
> Thanks for doing this. I'll have a look at the vhpiDepositPropagate thing at the weekend. Have you got a fork of cocotb that you're using for testing?

No, it's just vanilla. The commands from cocotb/cocotb@7f07ddac362423d701b7e90f05c1c9c295e5b1e8 work fine.

> Maybe try configuring with `--enable-asan` and see if that triggers.

Will do.

src/vhpi/vhpi-model.c
sean-anderson-seco: Yes

src/rt/model.c
sean-anderson-seco: OK, thanks for the explanation. I couldn't figure out how width differed from that calculation.

nickg (approved):
OK I'm happy to merge this now: it's clearly an improvement on the current state.

nickg:
I added `vhpiDepositPropagate` in 9b8e4be but I don't have access to another simulator to compare against at the moment. I tried cocotb but it failed with:

```
VHPI: vhpi_get_value expr=0x565078204ff0:{vhpiPortDeclK Name=Q} value_p=0x565078845ff8                                                                                                         
** Error: (init): invalid format 1 for object Q: expecting 23
    > dff.vhdl:11                              
    |                                          
 11 |   q: out std_logic);
    |   ^                                      
VHPI: vhpi_get_time time_p=0x7ffe3af7bf18 cycles=(nil)
     0.00ns ERROR    gpi                                VHPI Error level 3: invalid format 1 for object Q: expecting 23
                                                        FILE dff.vhdl:11
VHPI: vhpi_get_time time_p=0x7ffe3af7bf18 cycles=(nil)
     0.00ns ERROR    gpi                                VHPI: Size of m_binvalue.value.str was not large enough: req=-1 have=2 for type vhpiLogicVal

```

I probably made some mistake in the setup: I did `pip install cocotb` and then tried your command above in a clone of the current master.

sean-anderson-seco:
I see the same error.

I added the following debug to `vhpi_get_value` before the switch statement:

```c
VHPI_TRACE("type=%s (%s)", istr(type_name), type_kind_str(type_kind(base)));
```

and got `VHPI: vhpi_get_value type=STD_LOGIC (T_ENUM)`. The problem with this is that `is_well_known` expects a fully-qualified identifier, such as `IEEE.STD_LOGIC_1164.STD_LOGIC`.

nickg:
> I see the same error.
> 
> I added the following debug to `vhpi_get_value` before the switch statement:
> 
> ```c
> VHPI_TRACE("type=%s (%s)", istr(type_name), type_kind_str(type_kind(base)));
> ```
> 
> and got `VHPI: vhpi_get_value type=STD_LOGIC (T_ENUM)`. The problem with this is that `is_well_known` expects a fully-qualified identifier, such as `IEEE.STD_LOGIC_1164.STD_LOGIC`.

Thanks. I fixed this in 4168d0c but now I get a crash when cocotb calls `vhpi_get_value`: it looks like the passed in buffer address/size are bogus so it might be another `vhpi_user.h` incompatibility like you mentioned above.

> 
> This latter problem is because `vhpiCbLastKnownDeltaCycle` may itself create more delta cycles:
> 
> > h) If the next simulation cycle will be a delta cycle, the remainder of step h) is skipped. Otherwise, the following actions occur in the indicated order:
> > 
> > 1. Each registered and enabled `vhpiCbLastKnownDeltaCycle` and `vhpiCbRepLastKnownDeltaCycle` callback is executed. T_n is recalculated according to the rules of 14.7.5.1.
> > 2. If the next simulation cycle will be a delta cycle, the remainder of step h) is skipped.
> >    ...
> 
> A quick fix for that is something like
> 
> ```diff
> diff --git a/src/rt/model.c b/src/rt/model.c
> index cb7ce856..ce59763f 100644
> --- a/src/rt/model.c
> +++ b/src/rt/model.c
> @@ -3015,9 +3015,11 @@ static void model_cycle(rt_model_t *m)
>  
>     global_event(m, RT_END_OF_PROCESSES);
>  
> +   if (!m->next_is_delta)
> +      global_event(m, RT_LAST_KNOWN_DELTA_CYCLE);
> +
>     if (!m->next_is_delta) {
>        m->can_create_delta = false;
> -      global_event(m, RT_LAST_KNOWN_DELTA_CYCLE);
>  
>        // Run all postponed processes and event callbacks
>        workq_start(m->postponedq);
> ```

That change looks reasonable to me.

> 
> Also related is that callbacks can be used-after-free'd if they are free'd in another callback while they are already in a `workq`. I'm not sure what the best approach there is.

I guess we could either reference count them or store the freed callbacks in a quarantine area until there are known to be no references to them before actually freeing. I'll have a think about it.

nickg:
> Is there a general solution to identifiers not using fully-qualified names?

I can't actually remember why subtypes _aren't_ fully qualified. There's probably some latent bugs there as subprogram name-mangling could produce collisions. I've fixed that in 326ac99c8a.

cmarqu:
Is the general feeling here that it's worth to re-add NVC support in cocotb that was removed in https://github.com/cocotb/cocotb/pull/1694/files?
A pull request would likely be merged.
A follow-up PR could add NVC/main to cocotb's CI (with the `continue-on-error` flag set).

sean-anderson-seco:
Yes, I would like to re-add support for NVC. At the moment, only scalar signals/ports are supported, but I'm hoping to add arrays soon. This limits us to just the `dff` example for now.

PR #699, opened 2023-05-22
Fix coverage generation for unary operators

sean-anderson-seco:
I'm not quite sure if this is the right approach, but it seems simple enough.


src/lower.c
```diff
@@ -1921,15 +1921,16 @@ static void lower_logic_expr_coverage(lower_unit_t *lu, tree_t fcall,
...
    flags |= COV_FLAG_EXPR_STD_LOGIC;
 
-   assert(tree_params(fcall) == 2);
-
-   bool is_array[2] = {
-      type_is_array(type_param(tree_type(decl), 0)),
-      type_is_array(type_param(tree_type(decl), 1))
-   };
+   const unsigned nparams = tree_params(fcall);
+   assert(nparams > 0);
```
nickg, line 1925:
```suggestion
   if (nparams != 2)
      return;
```

Since we don't care about anything except two-argument calls.

Could you also delete the call to `fmt_loc` on line 1916? I think I left that in while I was debugging something a few weeks ago.

nickg (commented):
@Blebowski could you have a look too?

src/lower.c
```diff
@@ -1913,23 +1913,18 @@ static vcode_reg_t lower_logic_expr_coverage_for(tree_t fcall,
...
-      type_is_array(type_param(tree_type(decl), 0)),
-      type_is_array(type_param(tree_type(decl), 1))
-   };
+   if (tree_params(fcall) != 2)
+      return;
 
    // Skip arrays -> Matches behavior of VCS and Modelsim
-   if (is_array[0] || is_array[1])
+   if (type_is_array(type_param(tree_type(decl), 0)) ||
+       type_is_array(type_param(tree_type(decl), 0)))
```
nickg, line 1927:
```suggestion
       type_is_array(type_param(tree_type(decl), 1)))
```

Typo 0 -> 1. Or maybe it's cleaner with a loop:

```c
for (int i = 0; i < nparams; i++) {
   if (type_is_array(type_param(tree_type(decl), i)))
      return;
}
```

sean-anderson-seco: Whoops; missed that while copy-pasting.

nickg (approved):

Blebowski:
Hi, only one remark on my side. Could you please add simple test under `test/regress` ? (cover16)

Something that tries all variants of operator overload for `std_logic` on :
- LHS = 1 bit , RHS = vector
- LHS = vector, RHS = 1 bit
- LHS = vector, RHS = vector,
- unary op , argument is vector.

nvc should not emit coverage for any of these combinations, so it might be good to capture that in
regression test.

nickg (approved):

PR #745, opened 2023-07-26
src: Wrap tree_t into hdl_node_t for scopes and symbols.

Blebowski:
Hi,
take this as an idea proposal, not a serious pull-request.
When I was looking into adding implementation of simple `integer` variable in verilog, I ran accross lack
of vlog support in symbol table. I tried to wrap the references to underlying `tree_t` into a `hdl_node` union.
This-way, I hope I will be able to re-use implementation of symbols, scopes and nametab also for verilog
parsing. I would of course need to extend API in `names.h` by something like: `insert_vlog_name`.

I am sending this early up-front when no work is actually done (besides that wrapping). This is basically
just to ask whether it makes sense, or it is complete non-sense. I am not sure if having single the symbol
table for both VHDL and Verilog is a good idea.

### DIFF
--- a/src/names.c
+++ b/src/names.c
@@ -44,6 +44,11 @@ typedef enum {
    O_NAMED,
 } overload_state_t;
 
+typedef union {
+   tree_t      vhdl;
+   vlog_node_t vlog;
+} hdl_node_t;
+
 typedef A(tree_t) tree_list_t;
 typedef A(type_t) type_list_t;
 
@@ -66,7 +71,7 @@ struct _spec {
    spec_t      *next;
    spec_kind_t  kind;
    ident_t      ident;
-   tree_t       tree;
+   hdl_node_t   node;
    unsigned     matches;
 };
 
@@ -75,7 +80,7 @@ typedef enum {
 } visibility_t;
 
 typedef struct {
-   tree_t        tree;
+   hdl_node_t    node;
    scope_t      *origin;
    visibility_t  visibility : 16;
    tree_kind_t   kind : 16;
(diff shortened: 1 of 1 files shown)


nickg:
No, we should keep them separate. What's the benefit of combining them? The logic in `names.c` is already closely tied to the semantics of VHDL. If we add Verilog in there too it will be too complex to understand.

nickg:
For basic Verilog we shouldn't need such a complex symbol table anyway. A simple hash table should be enough like `vlog-sem.c` uses at the moment.

Blebowski:
Thanks. I was expecting similar answer, and I was worried of over-complicating it to the point it would become a mess.

However, maybe I can think of one benefit. If names and types are kept together, it may be possible to map VHDL types to System Verilog types. E.g. imagine a VHDL module with `enum` ports instantiated in System Verilog design. Mapping VHDL enum to System Verilog enum might be simpler to implement if the types are kept together and distinguished only by flags. Similarly, using e.g. constants from System Verilog package in a VHDL design and vice-versa.

The idea is very far-fetched though, so I am closing the MR.

nickg:
> If names and types are kept together, it may be possible to map VHDL types to System Verilog types. E.g. imagine a VHDL module with enum ports instantiated in System Verilog design. Mapping VHDL enum to System Verilog enum might be simpler to implement if the types are kept together and distinguished only by flags. Similarly, using e.g. constants from System Verilog package in a VHDL design and vice-versa.

I think if we were going to do that it would be better to generate a proxy object to wrap the imported SV package which would then have the equivalent VHDL types, etc.

PR #748, opened 2023-07-31
VHDL2019 assert api - partial

Blebowski:
Add partial implementation of: [LCS2016-050](http://www.eda-twiki.org/cgi-bin/view.cgi/P1076/LCS2016_050)

Still need to look at tests, some are failing.

Escaped identifiers with leading underscores are used in `std.env` to minimize "exposure" of the
implementation specific function and shared variable (instead of e.g. `assert_cnt`).


nickg:
This isn't the right way to implement it: the counters should be in C (e.g. in `src/rt/stdenv.c`) and incremented when `x_assert_fail` is called. The VHDL functions can then call into native code to read/clear the counters. Similar to how the directory and time functions in that file work.

Blebowski:
OK, I will give it some more work. So the counters will be incremented by `x_assert_fail`. To access the counters from `std.env.body`, I should then use the foreign function that then links agains the C implementation (similarly as in e.g. `file_line`)?

Blebowski:
OK, I see, similar to: https://github.com/nickg/nvc/commit/ea457d5de2771095f4fcf64419b13073d006c29e

Blebowski:
Now ready for review

lib/std.19/env-body.vhd
```diff
@@ -479,30 +479,40 @@ package body env is
     end function;
 
     impure function IsVhdlAssertFailed return boolean is
+        function impl (Level : severity_level) return natural;
+        attribute foreign of impl : function is "_std_env_get_vhdl_assert_count";
     begin
-        report "not implemented" severity failure;
+        return (impl(warning) > 0) or (impl(error) > 0) or (impl(failure) > 0);
```
nickg, line 485:
```suggestion
        IsVhdlAssertFailed(warning) or IsVhdlAssertFailed(error) or IsVhdlAssertFailed(failure);
```

lib/std.19/env-body.vhd
```diff
@@ -479,30 +479,40 @@ package body env is
...
-        report "not implemented" severity failure;
+        return (impl(Level) > 0);
     end function;
 
     impure function GetVhdlAssertCount return natural is
+        function impl (Level : severity_level) return natural;
+        attribute foreign of impl : function is "_std_env_get_vhdl_assert_count";
     begin
-        report "not implemented" severity failure;
+        return impl(warning) + impl(error) + impl(failure);
```
nickg, line 500:
```suggestion
        return GetVhdlAssertCount(warning) + GetVhdlAssertCount(error) + GetVhdlAssertCount(failure);
```

src/rt/model.c
```diff
@@ -111,6 +111,7 @@ typedef struct _rt_model {
    memblock_t        *memblocks;
    model_thread_t    *threads[MAX_THREADS];
    ptr_list_t         eventsigs;
+   rt_assert_cnts_t   assert_cnts;
```
nickg, line 114:
```suggestion
   unsigned   assert_cnts[SEVERITY_FAILURE + 1];
```

That way you don't need all the case statements.

test/regress/assert8.vhd
```diff
@@ -0,0 +1,55 @@
...
+            report "IsVhdlAssertFailed(F): "        & boolean'image(IsVhdlAssertFailed(failure));
+            report "************************************************************************";
+        end;
+    begin
+
+        print_asrt_state;
+
+        assert (False) report "First warning" severity warning;
+
+        print_asrt_state;
```
nickg, line 31: It's better to make the test self-checking and assert on the result of `GetVhdlAssertCount`, `IsVhdlAssertFailed`, etc. E.g.

```vhdl
assert GetVhdlAssertCount(warning) = 1;
assert IsVhdlAssertFailed;
```

lib/std.19/env-body.vhd
Blebowski: What is the benefit of these two changes besides reducing number of places in the code where
the run-time is called? Isn't it going to be slower (altough negligibly, but still) to execute when
the "wrapping" function is called first instead of directly the function that returns the counter?
Or will the optimizations take care of that?

test/regress/assert8.vhd
Blebowski: That is true, but still the test needs to have "gold" report since the asserts are expected to fail. Return
code will not distinguish correct/incorrect

lib/std.19/env-body.vhd
nickg: The overhead of the extra call is minimal and should be inlined anyway. This way avoid duplicating the symbol name / interface of the foreign function.

nickg (approved):

nickg:
I added the format string part.

PR #757, opened 2023-09-12
WIP: SDF parsing

Blebowski:
Hi,

This is initial MR for SDF parser. It is not ready for merge, I send it here just to get some feedback whether
you are OK with it, or whether you would drop it completely. It is almost orthogonal to the rest of the code base.

I keep the same structure of SDF objects as is for the tree/psl/vlog/type. I have extended the `object.c` code
to encode up to 8 different object types. Its just minor modification of shifts, not sure if it has some negative effect
on the rest of the codebase.

The parser uses YACC similarly as the Vlog stub that you have. I have started from the Vlog in sort of 
"copy and change" fashion. The lexer is almost untouched besides defining new mode and adding the
keywords. There is one small conflict if Vlog exists at the same time. For now, I have just commented
the vlog global vars. Eventually, I will need to resolve this, e.g. by adding prefix to the SDF parser.
AFAIK, this is the preffered way to do it once you have multiple generated parsers by YACC.
I didn't want to do it, because the code in `scan.c` relies on `yyval` and `yyloc` for proper lexing and
figuring out the source location. Having all the `loc_t` infrastructure seemed convenient.

As for the parsing, I use global context to track current cell, delay and delay_kind. I am not sure this
is the preffered way since it turns "context free grammar" into "context dependant parser". However,
SDF grammar is simple, and no recursions occur, so it should not matter. I first tried to keep no
context, and return all crated objects, but the `.y` file quickly became total mess.


nickg:
I'm not opposed to adding SDF in general, but I do think it would be better if it did something useful. E.g. ModelSim has a `-sdf` option to set VITAL generics from an SDF file. So we could have an elaboration option `--sdf=typ,file.sdf` or similar.

> I keep the same structure of SDF objects as is for the tree/psl/vlog/type.

I think it would be better in this case to parse it directly into plain C structs. The `object_t` infrastructure is mainly useful for automatic serialisation to disk and garbage collection of unused nodes, but I don't think we need either here. E.g.

```c
typedef struct {
   char *design;
   char *vendor;
   ...
   sdf_cell_t *cells;
} sdf_delayfile_t;

typedef struct {
   sdf_cell_t *next;
   char *celltype;
   ...
} sdf_cell_t;
```

Blebowski:
```
I'm not opposed to adding SDF in general, but I do think it would be better if it did something useful. E.g. ModelSim has a -sdf option to set VITAL generics from an SDF file. So we could have an elaboration option --sdf=typ,file.sdf or similar.
```

That is the plan, but first I need to at least get it parsed/dumped and somewhat checked. I don't expect to have it merged while "doing nothing". I want to put together at least basic Vital annotator and test against some free models and models of older Xilinx FPGA cells.

As for the controlling of the annotation, there is whole bunch of things to be configured, like the root hierarchy where to apply, corner selection, way to treat negative tchecks, disabling parts of tcheks, etc... It will most likely end up with small config file (similar to coverage exclusion, coverage spec files).

```
I think it would be better in this case to parse it directly into plain C structs. The object_t infrastructure is mainly useful for automatic serialisation to disk and garbage collection of unused nodes, but I don't think we need either here. E.g.
```

I am not sure. Parsing to structs may be faster, however, dumping the file onto the disk may be needed. I know Xcelium does this trick, where you pass SDF file only once, and it compiles it into a binary compiled form. Then, for next elaborations, you only pass the compiled SDF file and the elaboration and annotation are much faster. That would indicate compile/check is the most time consuming operation, rather than the annotation itself. Even simple SDF files are millions lines of code, so parsing it upon each elaboration may be a bottleneck. This feature helped me a lot during pre TO debug loops with Xcelium.

Either way, If I keep the API in `sdf-node.c` , the underlying representation of the nodes by structs or `object_t` should be easy to change.

Blebowski:
This reminded me a perfect article I read about SDF sign-off gate sims:

http://www.deepchip.com/items/0569-02.html

nickg:
> This reminded me a perfect article I read about SDF sign-off gate sims:
> 
> http://www.deepchip.com/items/0569-02.html

Thanks, it's an interesting article. I never used gate level simulation before when I used to do FPGA design.

Blebowski:
Its rather uncommon to do gate sims in FPGA devices. If something does not work, you simply generate new bit-stream. Commercially, I have seen it only in Aero-space design that needed to meet DO-254 certification.

The feature is rather used in ASICs, where it can help to discover bugs in the clock tree, reset tree, etc... It helps to reveal bugs, that cause your silicon to never come out of reset. And even with ASICs, since the devices got much more complex over the past years, SDF annotating full design with 100M+ gates is barely feasible. People still do gate-sims, but mostly zero delay to check e.g. ATPG pattern. Those simulate much faster if the simulator has parallel event processing (lot of events in single simulation time).

Still, I think SDF annotation is extremely usefull when you do smaller chips (lets say below 10-20M gates). Especially when you have heavily clock gated design, use non-reset flops to save area, or use other "non-standard" design techniques.

nickg:
@avelure mentioned they do VHDL gate-level simulation with back-annotation using the output from the Microsemi toolchain. It would be good if we could get that flow working here.

Blebowski:
OK, I will keep that in mind. @Avelure do you know where I could find the VITAL models of Microsemi / Microchip FPGA cells? I have tried to install latest Libero, and search through the installation for VHDL files that include `vital` keyword, but I can only find the Vital packages themselves, or some "dummy" examples.

I would expect, that after PnR i can write down the VHDL netlist and SDF. Then, If I compile Vital cell models + netlist, I can annotate the design.

avelure:
Microsemi changed to verilog libraries in the later Libero versions unfortunately. However we are using some older devices, the Proasic3 and RTAX, which are anyway not supported in the later versions. You need to install Libero 9.2 SP4. You will then find the device libraries with vital annotations in Designer\lib\vtl\95\[proasic3e.vhd|axcellerator.vhd]
Libero\designer tool can output a VHDL95 backannotated netlist with SDF.

Blebowski:
Perfect, thanks, that is exactly what I was looking for.

Blebowski:
Hi,

@nickg I don't really have time to work on this now. However, I have one question that came to my mind.

In VHDL, the SDF annotation is done by setting generics. In Verilog, it is done by modifying "specparam"
of specify blocks, or directly setting values in the Timing checks / IO delays.

For VHDL, you can usually do annotation by some simulator command like you mentioned above.
For Verilog though, you can also call `$sdfannotate` during the simulation itself (theoretically, at
non-zero time).

I was wondering, when to apply the values. To apply them, the design must be elaborated.
But in NVC, elaboration runs simultaneously with code-generation (Elab calls `elab_lower`).

I was thinking to put a hook into `elab_generic`, and do a lookup for the SDF value, and override
the generic there (similar to overriding the value by `-g ` command line).

I bet something similar will be eventually possible during Verilog elaboration at some point
during elaboration of timing check `vlog_node_t`.

This will however not work at all if the annotation were to be done during-time. I guess setting these
values at run-time, one would need to know exactly where (and how) did get the float constant representing
the delay / setup / hold translated. Do you think this is feasible ?

How would you approach this ?

nickg:
`elab_generic` is probably the right place to do it for VHDL. It would just detect if it's elaborating a VITAL model and then look up the value of the generic from the SDF data if provided. But I'd rather hold off doing this for now as I am planning to refactor a lot of the elaboration code.

Blebowski:
I don't think it is relevant at this point, I want to update the parser first.
I want to move to hand-written RDP as you recently did for Verilog.
It is better than generated one. Simpler to debug and maintain.
Also, I don't need to wrap my head around writing syntax productions 
without any shift/reduce reduce/reduce conflicts.

nickg:
> I want to move to hand-written RDP as you recently did for Verilog. It is better than generated one. Simpler to debug and maintain. Also, I don't need to wrap my head around writing syntax productions without any shift/reduce reduce/reduce conflicts.

Yes I always end up regretting using a parser generator for that reason. But the advantage is it's easier to get started. BTW I also used Bison for parsing VHDL until around 2013, that was a big mistake.

Blebowski:
Hi @nickg,

I have implemented almost whole RDP for the SDF according to SDF 4.0. I have copy-pasted the verilog parser
infrastructure and adjusted it slightly. I am slowly adding unit tests in this MR.

Since I have the parser complete, I was curious to try in, on a real SDF file. I use one from our recent tape-out.
The file is about 10 M lines of code and around 900 MB. It helped me to catch couple of real issues that unit tests
would not catch. However, I am running in trouble later in the file. When I introduce an error in the beginning of
the file, I am able to see the location, e.g.:
```
nvc -M 1024m -a regress/cover1.vhd -e --sdf tassic_top_Func_Osc_SignOff_Tc.sdf cover1 
** Error: unexpected ( while parsing delay file, expecting )
   > tassic_top_Func_Osc_SignOff_Tc.sdf:3
   |
 3 | (DESIGNs "tassic_top")
   | ^ this token was unexpected
```

But later the SDF parser crashes at:
```
nvc -M 1024m -a regress/cover1.vhd -e --sdf tassic_top_Func_Osc_SignOff_Tc.sdf cover1 
** Error: unexpected ( while parsing sdf cell, expecting )
   Note: this token was unexpected
```
without printing `loc_t`. I suspect this is due to `loc_t` being invalid. Looking at its definition,
the index of the first line can't fit into `first_line` bitfield. Having 20 bits for this information
will likely not be sufficient for decent size verilog netlist either.

Any clue on how to approach this ? I

nickg:
I don't think this is a problem. It's not crashing and it degrades gracefully to print the message without the error location. 20 bits of line number is over a million lines which can only be machine generated and so is unlikely to have syntax errors, or any syntax error can be caught with a smaller input.

Blebowski:
I took some smaller SDFs, and found couple of parsing bugs that popped out below 1M lines of code.

Now the file parses OK. 10.8 M lines of code parses in about 27 seconds on my machine.
Peak memory consumption is somewhere around 8GB. It would be definitely possible to optimize
it, but it would be pre-mature at this point. 

I will leave the `loc_t` out of the `sdf_node_t`. It probably does not make sense to fill-it in when the
location is anyway invalid for most of the nodes within the file.

Blebowski:
Hi @nickg,

I more-less finished the parser. I still have some TODOs to be done, and adding some more parsing tests on valid inputs.
I have not really tested invalid inputs as SDF is mostly machine generated format.

I hold the SDF cells via the `object_t` infrastructure.
Apart from holding each cell in an array under the root node, I hold the cell in one of two hash maps:
 - Hash by hierarchy to a cell index.
 - Hash by cell-name to a cell index if hierarchy is a wild-card (` ( INSTANCE * )`)
 
This is to facilitate an easy lookup when doing the annotation.
If I parse a cell that already exist in the hash table, I add items to it to keep single entry in the hash map.
This has a drawback that `dump` will not give 100 % equal SDF as the original one.
However, it will get equal SDF tree once parsed again. I don't think this is an issue.
I wanted to avoid holding a list for each entry in the hash-map.

I am still facing an issue with the `loc_t`. SDF standard says, that the cells should be processed
in order they are placed in the SDF file. So, if there are two entries for a module instance,
one via cell_name and wildcard, second via hierarchy, I will apply both during annotation.
Since I don't have `loc->first_line` I can't distinguish first one to apply.
Potentially, this could result in incorrectly annotated SDF file.
I guess this can be solved by keeping just a dummy line counter with each `S_CELL`.

I implemented simple serializer / deserializer that dumps SDF file into a "compiled SDF file".
The idea is that user will have two ways how to operate:
1. Precompile SDF file, use compiled SDF file for each next elaboration:
```
nvc -a tb_top.vhd dut_top.vhd
nvc -a my_sdf_file.sdf   --> Writes out my_sdf_file.sdf.compiled
nvc -e --sdf=my_sdf_file.sdf.compiled tb_top -r
```
2. Parse in-place during elab:
```
nvc -a tb_top.vhd dut_top.vhd -e --sdf=my_sdf_file.sdf tb_top -r
```
 
I tried to compare, and reading in the serialized version takes slightly less than half of the time
than parsing from text:
```
 nvc -M 8192m -a test/regress/cover1.vhd -e -V --sdf test/sdf_3v0_typ.sdf cover1
** Note: initialising [11ms +32588kB]
** Note: loading top-level unit [0ms +0kB]
** Note: analysed SDF file: test/sdf_3v0_typ.sdf [11256ms +5897304kB]
** Note: wrote compiled SDF file to: test/sdf_3v0_typ.sdf.compiled [9304ms +733392kB]
** Note: read compiled SDF file from: test/sdf_3v0_typ.sdf.compiled [5011ms +3083824kB]
** Note: elaborating design [2ms +0kB]
** Note: saving library [1ms +0kB]
** Note: code generation for 6 units [15ms +0kB]
** Note: linking shared library [189ms +0kB]
```

At the expense of spending additional 100 % time to dump `sdf.compiled` upon analyze time,
each next elaboration will take only half of the time to "load" the SDF.
When the SDF files are huge, this can be significant. 

Also, SDF files and Verilog netlists are often written out compressed with GZIP.
Eventually, I would like NVC to auto-detect (either by suffix, or by parsing the file),
that the SDF is GZIPED, and unzip it on the fly before the parsing.
This will make the benefit of the "compiled SDF" even larger, since the AOT compilation into
the "compiled SDF" format will include decrompression of the original SDF file.

Could you plas have a look at the current state and let me know your opinion.
I don't expect it to get merged, it is still "doing nothing", but I want to avoid going in
direction that is rejected later.

nickg:
I don't really have time to review all this at the moment. Perhaps the best way to move forward would be for me to extract the lexer changes and skeleton of the parser from this and commit them separately as these seem fairly uncontroversial and the rest of these changes can live in this PR for the time being. At least that would make rebasing easier.

> 
> I hold the SDF cells via the `object_t` infrastructure. Apart from holding each cell in an array under the root node, I hold the cell in one of two hash maps:

I'm still not convinced storing the AST is the best way to go here. It might be better to build an optimised database directly in the parser. I don't know enough about SDF to make a decision here so I need to spend time learning it but at the moment I'm quite busy with work/family things.

> 
> I am still facing an issue with the `loc_t`. SDF standard says, that the cells should be processed in order they are placed in the SDF file. So, if there are two entries for a module instance, one via cell_name and wildcard, second via hierarchy, I will apply both during annotation. Since I don't have `loc->first_line` I can't distinguish first one to apply. Potentially, this could result in incorrectly annotated SDF file. I guess this can be solved by keeping just a dummy line counter with each `S_CELL`.

The right way to do this is to store a sequence number with all the objects that need to be ordered (like e.g. `tree_pos` does for positional associations), rather than comparing source locations.

Blebowski:
> I don't really have time to review all this at the moment. Perhaps the best way to move forward would be for me to extract the lexer changes and skeleton of the parser from this and commit them separately as these seem fairly uncontroversial and the rest of these changes can live in this PR for the time being. At least that would make rebasing easier.

Understood. I will polish the remaining things in the parser, rebase and let you know.

> > I hold the SDF cells via the `object_t` infrastructure. Apart from holding each cell in an array under the root node, I hold the cell in one of two hash maps:
> 
> I'm still not convinced storing the AST is the best way to go here. It might be better to build an optimised database directly in the parser. I don't know enough about SDF to make a decision here so I need to spend time learning it but at the moment I'm quite busy with work/family things.

Likely that would be more optimal solution. SDF is all about performance (elaboration and simulation), and these days it is not feasible on bigger chips. Even parallel event queue within single time instance does not help as annotated clock tree skew will spread single clock edge into many time steps.

SDF nodes are not really "transformed / simplified / translated" in any manner. Its all about reading them, putting them into a "table" and then having a fast lookup when elaborating the design. So AST may be redundant. For me it seemed easy to
start with it as I was replicating the VHDL tree.

Still, each cell has arrays of timing checks, delays, that may have variable amount of members / values and optional
elements such as conditions or labels. So some form of object management is needed and `object_t` was convenient.

In my case I was happy to see that the parsing into memory took more than twice the loading time of the serialized AST,
otherwise the whole "precompilation" would make no sense.

I don't know how optimized database would look like. I don't have enough SW development experience to judge this.
Feel free to propose something.

Either-way there is no rush on this. I let it wait for the last year due to being really busy by work and family too.

> > I am still facing an issue with the `loc_t`. SDF standard says, that the cells should be processed in order they are placed in the SDF file. So, if there are two entries for a module instance, one via cell_name and wildcard, second via hierarchy, I will apply both during annotation. Since I don't have `loc->first_line` I can't distinguish first one to apply. Potentially, this could result in incorrectly annotated SDF file. I guess this can be solved by keeping just a dummy line counter with each `S_CELL`.
> 
> The right way to do this is to store a sequence number with all the objects that need to be ordered (like e.g. `tree_pos` does for positional associations), rather than comparing source locations.

Yeah that is the right way to do so.

Blebowski:
Looking at the SDF file I was testing with, there will definitely be a bottle neck in the annotation.
Each standard cell instance has couple of timing checks and io path delays. When elaborating
this instance, iterating its delays will not be a problem. Iterating timin checks may be bigger issue
since even a simple flop has dozens of them. So a fast lookup for matching timing check may
be needed.

A performance killer would be interconnect annotation. In my SDF file it is done by placing the
INTERCONNECT delay under a design top cell that is split accross many CELL constructs.
The interconnect path then contains full hierarchical path of the interconnect. So finding a
matching interconnect delay would mean iterating accross array of all interconnect delays...

src/lexer.l
```diff
@@ -542,8 +600,8 @@ UNION         ?i:union
 <VLOG>"inout"            { return tINOUT; }
 <VLOG>"reg"              { return tREG; }
 <VLOG>"always"           { return tALWAYS; }
-<VLOG>"posedge"          { return tPOSEDGE; }
-<VLOG>"negedge"          { return tNEGEDGE; }
+<VLOG>{POSEDGE}          { return tPOSEDGE; }
+<VLOG>{NEGEDGE}          { return tNEGEDGE; }
```
nickg, line 604: This isn't correct because it makes `posedge` and `negedge` case-insensitive in Verilog. You can leave it as it is and use `{POSEDGE}` and `{NEGEDGE}` for SDF only. I think you can also write `<SDF>?i:POSEDGE` directly in the place where it's used (not tested).

src/sdf/sdf-parse.c
```diff
@@ -0,0 +1,2739 @@
...
+   // te_spec ::=
+   //       ( TIMINGENV te_def { te_def } )
+
+   BEGIN("te spec");
+
+   consume(tLPAREN);
+   consume(tTIMINGENV);
+
+   // TODO: te_def should be obligatory at least once
+   while (scan(tLPAREN)) {
```
nickg, line 1592: This can be done by replacing the while loop with do-while (same in other places).

Blebowski:
Polished on my side, feel free to take over if you feel like it and when you have time.

The one remaining TODO can be solved later.

As for the AST alternatives, I think to get it right, there is the need to re-think the way annotation will be done.

I can think of following kinds:

VHDL:
- Setting VITAL generics based on entries in the SDF file.

Verilog:
- Cell delays, Timing checks and SpecParams do have directly corresponding HDL construct.
- Interconnect delays do not have such construct, so the delay will likely need to be split into port delays of hooked target cells.

As for the step where annotation is done, this can be:
- At elaboration time when `-sdf` option is passed.
- At run-time via `$sdfannotate`. This can be theoretically at non-zero time, so the code generated for the annotated unit would either need to recompile, or there would be direct deposit into some run-time changeable parameter (emited variable) that determines the delay / timing check value.

The approach with `elab_generic` callback could be extended towards `elab_timing_check` or `elab_port_delay` in Verilog.
Knowing generic / timing check / port delay hierarchy during elab, NVC would need to search fast for the matching item in the SDF file in some sort of hash table or database. I don't know how this would work for the interconnect delays as these are specified for pairs of module ports hooked to the same net.

Another way could be, that SDF annotation would be a separate step after elaboration or after code-gen. The hierarchy would be built, and then the SDF DB would be iterated as a list, and each item processed. A fast search for a HDL construct in the elaborated / code generated design would be needed then. If this was supposed to be done at run-time, then the SDF values would need to be deposited into already generated / running code.

The commercial tools do the SDF annotation at start of run-time as separate step without any recompilation.

Blebowski:
Just an update:

I was wrong in the end of previous post. One of the commercial simulators I was referencing was
doing the SDF annotation at elaboration time, not at run-time. Even if `$sdf_annotate` was present
in the code, it annotated the design "right-away". The SDF file argument then must be elaboration
time constant. I think this would be enough in NVC too. There is no real use-case for annotating
in the middle of simulation.

nickg:
I extracted the parser/lexer changes from this and pushed it in 4f7e0594a. To rebase this branch you'll probably need to do `git checkout --theirs src/sdf/sdf-parse.c` rather than fixing the conflicts by hand.

Blebowski:
Thanks. I am not sure it makes sense rebasing though. I will give it a try and see based on amount of trouble this causes.

The only "extra" thing kept there is the AST + guts of the unit tests. You mentioned the AST may be redundant, and
more I think of it, you are right. Maybe keeping flat array or linked list of structs where each struct holds single item
to annotate would be better solution. The "cell" level could be completely dropped.

I would be curious about what do you think about the annotation process ?
Do you think it is feasible to do it like so:
```
walk tree, elaborate, only hold elaborated tree.
for <annotation_item> in <all_sdf_items_of_a_file>:
    visit tree / vlog item based on hier of <annotation_item> -> Need fast way of finding tree / vlog based on hier name.
    override parameter / generic of tree / vlog
walk tree again do lower.
```
or you rather see this as:
```
walk tree, in elaborate:
    When Generic / Parameter / Timing check / Specify block is visited, then
    search the SDF file for matching SDF item -> Needs a fast way of finding the SDF item based on hier.
    do lower of the instance within the elab.
```
Its about deciding if SDF items are to be kept flat, or looked-up based on hierarchy.
The lookup may be tricky since some SDF items (e.g. interconnect delays) have multiple hiers.
The interconnect delays hold source port and load port that are connected to a single net.
These should be somehow split to port delays on the source / load, see 32.4.4 in SV2012 LRM.
So if lookup is done e.g. by source hier, one still needs to find all load cells...

PR #767, opened 2023-10-06
src: Rename conditional compilation tokens

Blebowski:
SDF has `COND` and `CONDELSE` keywords. I am trying to keep the naming of tokens to be `t<KEYWORD>`.

Thus I run into conflicts with conditional analysis tokens. I renamed them to be `tCND...` instead of `tCOND...`.

Thus `CONDELSE` from SDF does not conflict with `\`else` conditional analysis identifier.


nickg:
I think you can just reuse those tokens? At the moment the scanner will only generate them in the VHDL (default) state.

Blebowski:
> I think you can just reuse those tokens? At the moment the scanner will only generate them in the VHDL (default) state.

OK, will do so then.

PR #771, opened 2023-10-15
src: FSM state coverage.

Blebowski:
Hi,

this is an implementation of FSM state coverage. When `fsm-state` coverage type is enabled,
NVC tracks coverage of visited literals of signals with enum type. By default, all user defined
enums are considered as FSMs. Further, there are options how to exclude / include particular
types to be considered as FSMs.

Implementation-wise, each state is tracked in single `coverage_tag_t`. This is to propagate
state name to the generated data, and guarantee unique `coverage_tag_t` hierarchies for
merging. Compared to the solution with multiple states packed to single tag, the solution is
simpler. Packing and unpacking does not need to be done, and whole report generation
infrastructure is reused, since "state bin" is tracked for each coverage tag by
`COVER_FLAG_STATE`.

The memory overhead of a new tag for each state is negligible, considering that even very
complex HW designs have at most 100s of FSMs, giving 1000s of states.

The report generation groups states of a single FSM to a table, and treats each state as
a coverage bin. This is little different that other coverage types where bins are flags
within single coverage tag. In FSM state coverage, each tag creates a bin, and multiple
tags cause single table with bins to be drawn in the table. In couple of places this requires
little special handling, but the overall feeling of the output HTML table is worth it.

The work is ready for review apart from one issue:
When I specify coverage specification file, NVC crashes when collecting FSM coverage.
This can be seen in `cover19` test. The reason is, that the scope for the signal is
not emitted, because coverage spec file does not contain the FSM signal hierarchy.

I still need to figure out how to fix this bug.


src/jit/jit-irgen.c
```diff
@@ -3433,6 +3433,18 @@ static void irgen_op_cover_toggle(jit_irgen_t *g, int op)
    macro_exit(g, JIT_EXIT_COVER_TOGGLE);
 }
 
+static void irgen_op_cover_state(jit_irgen_t *g, int op)
+{
+   jit_value_t shared = irgen_get_arg(g, op, 0);
+
+   uint32_t tag = vcode_get_tag(op);
+   jit_value_t mem = jit_addr_from_cover_tag(tag);
```
nickg, line 3441: Please change this to pass the tag number rather than a pointer to the memory (see 140d8d648e). The memory might get reallocated later before the callback fires.

src/rt/cover.c
```diff
@@ -1164,6 +1239,42 @@ void x_cover_setup_toggle_cb(sig_shared_t *ss, int32_t tag)
...
+      break;
+   case 2:
+      tag_index = *((uint16_t*)signal_value(s));
+      break;
+   case 4:
+      tag_index = *((uint32_t*)signal_value(s));
+      break;
+   default:
+      fatal("Unsupported signal size: %d\n", size);
+   }
```
nickg, line 1260: Use the `FOR_ALL_SIZES` macro. Something like:

```c
#define READ_STATE(type) tag_index = *((type *)signal_value(s));
FOR_ALL_SIZES(size, READ_STATE);
```

src/rt/cover.c
```diff
@@ -1164,6 +1239,42 @@ void x_cover_setup_toggle_cb(sig_shared_t *ss, int32_t tag)
...
+
+   uint32_t *mask = ((uint32_t*) user) + tag_index;
```
nickg, line 1262: See the earlier comment about passing the tag number rather than the pointer to the callback and then change this to `get_cover_counter(m, tag_index)`.

Blebowski:
Hi, @nickg now this is ready for review.

I have applied the fix you proposed, and pass tag index instead of pointer. I do that for JIT op and in run-time code too.

The other problem is fixed too.

src/lower.c
```diff
@@ -1763,6 +1763,38 @@ static void lower_toggle_coverage(lower_unit_t *lu, tree_t decl)
...
+   cover_push_scope(lu->cover, decl);
+
+   int hops = 0;
+   vcode_var_t var = lower_search_vcode_obj(decl, lu, &hops);
+   assert(var != VCODE_INVALID_VAR);
+
+   // Add single coverage tag per enum literal. This is to track literal string
+   // in the identifier of the coverage tag.
+   for (int i = 0; i < type_enum_literals(type); i++) {
+      tree_t literal = type_enum_literal(type, i);
```
nickg, line 1783: I don't think this works if `type` is a subtype of an enum. It should be like:

```c
type_t base = type_base_recur(type);
const int nlits = type_enum_literals(type);
for (int i = 0; i < nlits; i++) {
   tree_t literal = type_enum_literal(type, i);
```

Or perhaps just restrict the coverage items to the literals in the subtype:

```c
int64_t low, high;
if (!folded_bounds(range_of(type, 0), &low, &high))
   return;

type_t base = type_base_recur(type);
for (int i = low; i < high; i++) {
   tree_t literal = type_enum_literal(type, i);
```

src/rt/cover.c
```diff
@@ -310,6 +324,51 @@ bool cover_skip_array_toggle(cover_tagging_t *tagging, int a_size)
...
+   // Type should be included
+   if (spc)
+      for (int i = 0; i < spc->fsm_type_include.count; i++)
+         if (ident_glob(name, AGET(spc->fsm_type_include, i), -1)) {
+#ifdef COVER_DEBUG_EMIT
+            printf("Cover emit: True, fsm-type (Type: %s, Pattern: %s)\n",
+                  istr(name), AGET(spc->fsm_type_include, i));
+#endif
+            return false;
+         }
```
nickg, line 352:
```suggestion
   if (spc) {
      for (int i = 0; i < spc->fsm_type_include.count; i++)
         if (ident_glob(name, AGET(spc->fsm_type_include, i), -1)) {
#ifdef COVER_DEBUG_EMIT
            printf("Cover emit: True, fsm-type (Type: %s, Pattern: %s)\n",
                  istr(name), AGET(spc->fsm_type_include, i));
#endif
            return false;
         }
   }
```

src/rt/cover.c
```diff
@@ -310,6 +324,51 @@ bool cover_skip_array_toggle(cover_tagging_t *tagging, int a_size)
...
+   // Type should not be included
+   if (spc)
+      for (int i = 0; i < spc->fsm_type_exclude.count; i++)
+         if (ident_glob(name, AGET(spc->fsm_type_exclude, i), -1)) {
+#ifdef COVER_DEBUG_EMIT
+            printf("Cover emit: False, fsm-type (Type: %s, Pattern: %s)\n",
+                   istr(name), AGET(spc->fsm_type_exclude, i));
+#endif
+            return true;
+         }
```
nickg, line 367:
```suggestion
   if (spc) {
      for (int i = 0; i < spc->fsm_type_exclude.count; i++)
         if (ident_glob(name, AGET(spc->fsm_type_exclude, i), -1)) {
#ifdef COVER_DEBUG_EMIT
            printf("Cover emit: False, fsm-type (Type: %s, Pattern: %s)\n",
                   istr(name), AGET(spc->fsm_type_exclude, i));
#endif
            return true;
         }
     }
```

src/lower.c
Blebowski: I see, using `type_base_recur` will help here for the subtype.

But the second counter-proposal won't work either. If I don't emit tags for literals starting from first (translated to 0x0), the run-time search for pointer to coverage data won't work correctly. To get it working, we would need to store index of the first literal in the tag, and then subtract it from actual enum value in the runtime code.

nickg: > 
> But the second counter-proposal won't work either. If I don't emit tags for literals starting from first (translated to 0x0), the run-time search for pointer to coverage data won't work correctly. To get it working, we would need to store index of the first literal in the tag, and then subtract it from actual enum value in the runtime code.

OK let's just leave that for now.

Blebowski: When you wrote it, I was half-way through implementation. I tried to give it a shot. 

I keep the `low` as an argument of the op, and I subtract from the index of the first tag upon callback setup.
This pre-compensates the tag index by a negative offset. Then, run-time CB adds the offset back, and gets
correct index of counter corresponding to first tag for given sub-type.

Blebowski: Note that all this heavily relies on the fact that enums are numbered from zero in the generated code, and that sub-type enums have the same encoding as corresponding base type enums. I think it is reasonable to rely on that.

The current "addressing" of tag would completely fall apart if there would be a possibility to encode enum with non-monotonic values (e.g. with C/System Verilog like base-type). Luckily VHDL does not have this.

nickg (approved):

PR #777, opened 2023-10-22
src, test: Refactoring of cover. 

Blebowski:
Rename "tag" to "item", and "tagging" to "ctx".

### DIFF
--- a/src/cgen.c
+++ b/src/cgen.c
@@ -53,7 +53,7 @@ typedef struct {
    char            *obj_path;
    char            *module_name;
    unsigned         index;
-   cover_tagging_t *cover;
+   cover_data_t    *cover;
    llvm_obj_t      *obj;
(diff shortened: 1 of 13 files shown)


nickg:
I don't like the name `cover_ctx_t` very much - can we rename it to `cover_data_t` instead?

Blebowski:
Ok, not an issue. How would you then name local variables of `cover_data_t` type in `cover.c` ?
Something like: `data` (seems to generic and confusing). Maybe `cover_data` or simply `cover`
(as in rest of the code-base apart from `cover.c`) ? I would then unify it also in `model.c` , `nvc.c`
and other places.

nickg:
Maybe `data` in `cover.c` and `cover` everywhere else.

Blebowski:
Done

nickg (approved):

PR #780, opened 2023-10-23
Split src/rt/cover.c into multiple files

nickg:
The code coverage implementation is getting quite large now and the database, runtime, and report generation code is all mixed together. I tried to split it into several logically separate files under a new `src/cov/` sub-directory. The original `src/rt/cover.c` now only contains the runtime bits. @Blebowski what do you think?


src/cov/cov-data.c
```diff
@@ -0,0 +1,823 @@
...
+#include <string.h>
+#include <stdlib.h>
+#include <limits.h>
+#include <libgen.h>
+
+//#define COVER_DEBUG_EMIT
+//#define COVER_DEBUG_DUMP
+//#define COVER_DEBUG_SCOPE
+//#define COVER_DEBUG_MERGE
+//#define COVER_DEBUG_CALLBACK
```
Blebowski, line 39: I think `COVER_DEBUG_EXCLUDE` should go to `cover_exclude.c` and `COVER_DEBUG_CALLBACK` to `rt/cover.c`

Blebowski (changes requested):
Hi, I was wondering about similar thing, thansk for doing the work. See only the one comment I have.

Issue #823, opened 2024-01-05
Support merging covdbs with different top-level architectures


sean-anderson-seco:
At the moment, it is only possible to merge covdbs which share the same top-level architecture. Merging coverage databases with different top-level architectures results in only the coverage from the first database being kept.

However, it is common to instantiate architectures in a hierarchy. This can result in misleading coverage results. For example, it could appear that `A.B` did not have sufficient coverage, even though there is a separate test for `B` which includes paths not covered in the test of `A.B`.

We should add support for merging covdbs with different top-level architectures, allowing a coverage summary to be created and easing identification of areas not covered under any test.

Blebowski:
Hi,

it always depends on what you are intending to cover. Sometimes, it is desirable to target every item for the given hierarchy.
If you have single entity instantiated multiple times (e.g. a memory instantiated multiple times via `generate`), you
would typically want to cover access to each such memory, so merging all the data for the memory entity from all its
instantiations will actually hide some of the uncovered data. When you have a TB with regression that achieves a coverage
for each such instantiation, you can be more-less sure, that once you replace your DUT with gate-level netlist, you will
catch cases where synthesizer optimized away part of your design. In other words, you check how much is your
"logic" covered.

Commercial tools support merging the coverage for each source file (instead of by hierarchy/logic). It would be
definitely good to add this type of coverage too. The report would then contain list of source files, instead of hierarchy.

What you are reffering to, is somehow mixture of both approaches. To have "per-hierarchy" coverage, but with option
to "add coverage" to a sub-block from a different TB that instantiates the same sub-block. E.g. if you have a top level
TB for an IP, but the IP has a sub-block whose logic is very hard to cover in the top level TB. So, you create a unit-test
for such sub-block, and you would like to have the coverage data for the sub-block from this unit test merged into the
hierarchy coverage of whole IP from the top-level TB.

I would not do that somehow automatically, otherwise there is the aliasing/hiding danger as described above. But,
I would maybe extend the `-c` command with an option like:
```
--fold=<entity_name>
```
Such option would "fold" together all the coverage that is under the `entity_name` block. Should it fold together
also sub-entities of `<entity_name>` ? I would say yes...

So, to sum it up, I think two things would be good:

1. A "per source file" merging mode (as oppose to "per hierarchy" mode).
2. Option to "fold" together coverage for some entity from different hierarchies when operating in the "per hierarchy" mode.

sean-anderson-seco:
[...]

PR #828, opened 2024-01-11
Cov different top hierarchies

Blebowski:
Hi,

this MR is follow-up of issue by @sean-anderson-seco  :
https://github.com/nickg/nvc/issues/823

It handles code coverage report generation when there are multiple scope children in the root
coverage scope. Previously, only first scope was reported.

### DIFF
--- a/src/cov/cov-report.c
+++ b/src/cov/cov-report.c
@@ -321,8 +321,9 @@ static void cover_print_file_and_inst(FILE *f, cover_report_ctx_t *ctx, cover_sc
 
    cover_file_t *src = cover_file(&(s->loc));
    fprintf(f, "<h2 style=\"margin-left: " MARGIN_LEFT ";\">\n");
-   fprintf(f, "   File:&nbsp; <a href=\"../../%s\">../../%s</a>\n",
-               src->name, src->name);
+   if (src != NULL)
+      fprintf(f, "   File:&nbsp; <a href=\"../../%s\">../../%s</a>\n",
+                  src->name, src->name);
    fprintf(f, "</h2>\n\n");
 }
 
@@ -1178,9 +1179,14 @@ static void cover_report_scope(cover_report_ctx_t *ctx,
    for (int i = 0; i < s->items.count; i++) {
       cover_item_t *item = &(s->items.items[i]);
 
-      cover_file_t *f_src = cover_file(&(item->loc));
-      if (f_src == NULL)
+      loc_t *loc = &(item->loc);
+      cover_file_t *f_src = cover_file(loc);
+      if (f_src == NULL) {
+         warnf("Could not locate source file: %s that NVC used to collect "
+               "coverage for: %s. Dropping coverage for this hierarchy/item "
+               "from coverage report." , loc_file_str(loc), istr(s->hier));
          continue;
+      }
 
       cover_line_t *line = &(f_src->lines[item->loc.first_line-1]);
 
--- a/src/cov/cov-report.c
+++ b/src/cov/cov-report.c
@@ -113,6 +113,8 @@ typedef enum {
 
 static cover_file_t  *files;
 
+#define COV_RPT_TITLE "NVC code coverage report"
+
 static void cover_report_children(cover_report_ctx_t *ctx,
                                   cover_scope_t *s, const char *dir,
                                   FILE *summf, int *skipped);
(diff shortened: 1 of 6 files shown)


src/cov/cov-report.c
```diff
@@ -284,51 +282,51 @@ static void cover_print_html_header(FILE *f, cover_report_ctx_t *ctx, bool top,
...
-   else
-      fprintf(f, "   Instance:");
+   fprintf(f, "   Instance:&nbsp;%s\n", istr(s->hier));
    fprintf(f, "</h2>\n\n");
 
    cover_file_t *src = cover_file(&(s->loc));
    fprintf(f, "<h2 style=\"margin-left: " MARGIN_LEFT ";\">\n");
-   if (!top)
-      fprintf(f, "   File:&nbsp; <a href=\"../../%s\">../../%s</a>\n",
+   fprintf(f, "   File:&nbsp; <a href=\"../../%s\">../../%s</a>\n",
```
sean-anderson-seco, line 324: I get a segfault here if NVC can't find the source file. This can occur if e.g. the working directory is different from when the architecture was run.

sean-anderson-seco:
I think the man page needs an update as well.

---

Thanks for this. It's definitely nicer than my previous script:

```html
#!/usr/bin/env bash

TITLE="NVC Coverage Report Index"

cat << EOF
<!DOCTYPE html>
<html lang="en">
<head>
	<meta charset="UTF-8">
	<title>${TITLE}</title>
</head>
<body>
	<h1>${TITLE}</h1>
	<ul>
EOF

for module in "$@"; do
	echo "<li><a href=\"${module}/index.html\">${module}</a></li>"
done

cat << EOF
	</ul>
</body>
</html>
EOF
```

src/cov/cov-report.c
Blebowski: Should be fixed now.

Blebowski:
> I think the man page needs an update as well.

At what place ?

src/cov/cov-report.c
```diff
@@ -284,51 +282,52 @@ static void cover_print_html_header(FILE *f, cover_report_ctx_t *ctx, bool top,
...
-
    fprintf(f, "<header>");
-   va_start(ap, title);
-   vfprintf(f, title, ap);
-   va_end(ap);
+   fprintf(f, COV_RPT_TITLE "\n");
    fprintf(f, "</header>\n\n");
+}
+
+static void cover_print_navigation_tree(FILE *f, cover_report_ctx_t *ctx, cover_scope_t *s, ...)
```
nickg, line 290:
```suggestion
static void cover_print_navigation_tree(FILE *f, cover_report_ctx_t *ctx, cover_scope_t *s)
```

Doesn't need to be a variadic function.

src/cov/cov-report.c
sean-anderson-seco: Thanks. I do see a bunch of warnings (around 2-3 per module) about the source file being null. Is that expected?

sean-anderson-seco:
> > I think the man page needs an update as well.
> 
> At what place ?

In the code coverage section. I think it would be nice to add an example with multiple coverage reports. And a comment on `--merge` would be good too, since right now it's not obvious how it works. Speaking of which, with this PR does `--merge` work if you specify unrelated covdbs? E.g. can you do `nvc -c --merge foo.covdb foo-first.covdb foo-second.covdb bar.covdb`? Or is it the same as before?

src/cov/cov-report.c
nickg: Perhaps needs to be guarded with `if (!loc_invalid_p(&(s->loc)) {` to make sure it has a valid location. Every node should really have a location set though.

Blebowski: The `loc_invalid_p` is called inside of `cover_file`. I have added a warning in `cover_report_scope` that fires when
`cover_file` returns NULL. Reporting is skipped for such scope, so I think it is good to let user know something is
missing from the generated report.

@sean-anderson-seco , I can't reproduce your case. I have tried to call the `nvc -c` from different folders that I used to run `nvc -a` with `covdb` path adapted accrodingly, but it works in my case. I tried deleting the original source file, and
the warning fires then.

nickg: It must be that the `tree_t t` argument to `cover_push_scope()` doesn't have a location set. @sean-anderson-seco can you work out what kind of node it is from the scope name? 

Anyway I don't think it's caused by the changes here.

nickg (approved):

Issue #788, opened 2023-11-02
Coverage: Source of covered / uncovered data


Blebowski:
When doing code coverage, there is no way how to find out in the generated report the original coverage
database that lead to the coverage item being covered or uncovered. This would be a nice feature.

This is related to adding `--name` option that would uniquely distinguish the elaborations of otherwise
equal design. We can either keep within `covdb` the path to the `covdb` itself, or keep the `--name` to distinguish
individual "runs".

Then, upon merging of coverage, the original path of all `covdb` that were merged would be placed into the `covdb`.
There would be a list of paths in the beginning of the `covdb` that would indicate all the DBs that were used to
get this final "merged" one.

During merging, we always know the `covdb` that is being currently "merged" into the "old" `covdb`. Thus, we could
keep per-tag mask of `covdb`s that lead to the tag being covered / uncovered. Only drawback is, that the number of
databases is not limited, so, the "per-tag" mask of covdbs would have variable length depending on the number of
covdbs. Or, we can support only e.g. up to 64 covdbs, but I would like this to work for unlimited number of `covdbs`.

Blebowski:
Hi,

I am trying to rethink how to implement this, and I come no an interesting limitation. Since branch, expression and toggle coverage encode multiple "bins" in single coverage item, if you merge such items together, it is very difficult to track covdb
where each of the bins was covered. It would essentially lead to tracking each of the masks separately, which ends up
with up to 4 variable length arrays. Simply, it is starting to seem messy...

It seems that the design decision to encode multiple "bins" into a single `cover_item_t` indeed makes the things more complicated than it is necessary. I am thinking to refactor this, and keep single `cover_item_t` for a single bin. The `num` attribute would then give how many follow-up `cover_item_t` correspond to the same "original coverage item". 

Thus, each branch with if/else and each toggle would be split into two coverage items (0/1, true/false). Each binary expression
would be split into up-to 4 coverage items (01,10,11,00). Anyway this is already done for "case" branch coverage where number
of bins is not limited, thus it is not possible to encode it into single item. Similarly, for FSM coverage, each state is encoded in its own `cover_item_t`, and multiple `cover_item_t` are aggregated into a single table during report generation.

With this change, the run-time data for each "bin" / `cover_item_t` would become a counter (the same as for statements), so the code for merging and converting the `cover_item_t` to `cover_chain_t` would simplify.

[...]

PR #865, opened 2024-03-08
WIP: Temporary coverage rework.

Blebowski:
Hi,

this is the MR you were asking about in the conversation. But it is in a very "raw" state...

The ideas are following:

1. Make single `cover_item_t` for a single coverage bin. Convert all the coverage run-time
data to counters (since single `cover_item_t` will represent single bin, no masking and shifting
is needed). This will simplify handling of coverage items, coverage merging, etc. Maybe VCODE
OPs could even be reworked to have only a single VCODE OP type (coverage OP) with some
sort of sub-kind that. The lower logic would change to emit new block that is conditionally
executed/skipped if the given condition/branch/expression is evaluated to true. The execution
code for each Coverage op will be just adding +1, so no complex handling will be needed in
the JIT layer for different coverage OPs. Basically everything will be the same as Statement
coverage is now. The name of the "BIN" would then be suffixed to the name of the emited
`cover_item_t`. The coverage exclude logic will then no need to specify any bins, everything
will be simply encoded in the name of the coverage item. This is where the user interface
would change slightly.

2. Consecutive "N" coverage items (that are now encoded in single item) will be still reported
in a single "table" in the report. To do that, one would need to identify how many consecutive
items belong to "the same construct". The `num` attribute of first such item can be used to
identify how many next consecutive items are to be grouped. It is already done like that for
FSM coverage. The `num` identifies all cover items to be put into single table since they
belong to single FSM. This can be generalized for all coverage items.

3. Simplify interface between `lower` and `cover` parts. Move all the logic that emits the
coverage items into `cover` (e.g. as is now done for the toggle coverage). The `cover_add_item`
would change to `cover_add_items` and return pointer to first coverage item added. Since
`num` would contain number of consecutive items, it would be simple to iterate these items
in `lower` and create the branching logic needed to execute/skip the VCODE cover op.
All the API functions like: `cover_inc_array_depth`, `cover_dec_array_depth`, `cover_skip_array_toggle`
`cover_skip_type_state` and `cover_get_std_log_expr_flags` would be now only local to
`cover` part.

Possible implications:
- Larger COVDB (2 x more toggle items, 2 x more branch items, 2 - 4 x more expression items)
- Not sure on the performance impact. Might be faster or slower, no idea.
- Simpler way of implementing https://github.com/nickg/nvc/issues/788. Additional member of
  `cover_item_t` could provide "source DB" for each BIN, and it could be easier to achieve
  "Bin A was covered in coverage DB X". This can be usefull to see that corner-case tests
  were really the ones that covered certain part of the code.

### DIFF
--- a/src/cov/cov-api.h
+++ b/src/cov/cov-api.h
@@ -235,16 +235,12 @@ void cover_export_cobertura(cover_data_t *data, FILE *f,
 void cover_push_scope(cover_data_t *data, tree_t t);
 void cover_pop_scope(cover_data_t *data);
 
-void cover_inc_array_depth(cover_data_t *data);
-void cover_dec_array_depth(cover_data_t *data);
-
 bool cover_is_stmt(tree_t t);
-bool cover_skip_array_toggle(cover_data_t *data, int a_size);
 bool cover_skip_type_state(cover_data_t *data, type_t type);
 
 unsigned cover_get_std_log_expr_flags(tree_t decl);
 
-cover_item_t *cover_add_item(cover_data_t *data, object_t *obj, ident_t suffix,
-                             cover_item_kind_t kind, uint32_t flags);
+cover_item_t *cover_add_items(cover_data_t *data, object_t *obj,
+                              cover_item_kind_t kind);
 
 #endif   // _COV_API_H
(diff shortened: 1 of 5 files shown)


nickg:
I split out just the branch coverage changes from this in 08505635e929d3.

Blebowski:
Thanks, doing this step-by-step is much better approach than all at once.

Blebowski:
I think it makes sense to close this MR, throw away the branch as dead and do the changes step-by-step.
I tried to give it couple of hours and I got lost. I ran into the pitfall of trying to fix everything at once...

PR #896, opened 2024-05-30
Fix basic_identifier

typingArtist:
In the LRM (VHDL-2008 at least) basic identifiers are defined as:

```
 basic_identifier ::= letter { [ underline ] letter_or_digit }
```

The LRM defines brackets `[]` as optional elements (regex question mark `?`) and braces `{}` as optional repeated (zero or more, regex asterisk `*`).

For underlines, this means that an identifier must never start with it, end with it, and there must never be more than one underscore in sequence. Before this PR, nvc doesn’t parse basic identifiers according to that definition but permits underscores in any place of an identifier, except for the first position. This patch checks for the proper behavior and also fixes the issue in the parser.

With just 47d0963 applied, `make check` fails. 62c2c0a fixes the issue. Neither verilog nor tcl were enabled, so a couple of unrelated tests have not been run.

While regenerating `test/dist.mk`, a lot of missing entries show up, but for atomicity of the patch, only the new dependency was added.

### DIFF
--- a/src/lexer.l
+++ b/src/lexer.l
@@ -94,7 +94,7 @@ extern yylval_t yylval;
 
 LOWER           [a-z\xdf-\xf6\xf8-\xff]
 UPPER           [A-Z\xc0-\xd6\xd8-\xde]
-VHDL_ID         ({LOWER}|{UPPER})({LOWER}|{UPPER}|[_0-9])*
+VHDL_ID         ({LOWER}|{UPPER})(_?({LOWER}|{UPPER}|[0-9]))*
 EXID            \\([^\\\n]|\\\\)*\\
 VLOG_ID         [a-zA-Z_]([a-zA-Z0-9_$])*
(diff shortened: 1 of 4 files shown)


nickg:
> While regenerating test/dist.mk, a lot of missing entries show up, but for atomicity of the patch, only the new dependency was added.

I normally only regenerate that when I'm about to make a release. It's not necessary to keep it always in sync.

nickg:
I extended this a bit in 8fd2b74 so it now prints a more helpful error message:

```
** Error: a VHDL basic identifier must not start or end with an underscore, or contain 
          multiple consecutive underscores
   > ../test/parse/basic_identifier.vhd:1
   |
 1 | entity ee_ is end entity; -- ERROR expecting is (after ee)
   |        ^^^
   |
   = Help: IEEE Std 1076-2008 section 15.4.2 "Basic identifiers"
```

PR #898, opened 2024-06-13
Add vlog celldefine

Blebowski:


nickg:
I think it would be better to ignore the celldefine directive for now (so the lexer should just skip over it) since we're not going to do anything with it for the foreseeable future.

nickg (approved):

PR #919, opened 2024-07-19
Cover saturate toggle

Blebowski:
Hi,

this is a MR with some left-overs from previous refactor:

- Couple of bug-fixes that creeped in and were un-recognized since they are related to Coverage report printing that is not checked.
- Removed `flag` from `cover_pair_t` -> Redundant now
- Added run-time saturation also for Toggle Items
- Optimizes toggle items in cases when `count-from-undefined` and `count-from-to-z` are enabled.
- Slightly improves coverage report for PSL assertions
- Added Cover item data into the Coverage report (now that it means how many times an item was covered).
- Added a test for saturation during merging. Testing Saturation on increments during run-time is excessively long.

### DIFF
--- a/src/cov/cov-data.c
+++ b/src/cov/cov-data.c
@@ -666,13 +666,17 @@ cover_item_t *cover_add_items_for(cover_data_t *data, object_t *obj,
 
 static void cover_merge_one_item(cover_item_t *item, int32_t data)
 {
+   int32_t inc;
+
    switch (item->kind) {
    case COV_ITEM_STMT:
    case COV_ITEM_FUNCTIONAL:
    case COV_ITEM_BRANCH:
    case COV_ITEM_STATE:
    case COV_ITEM_EXPRESSION:
-      item->data += data;
+      inc = item->data + data;
+      if (likely(inc > item->data))
+         item->data = inc;
       break;
 
    // Highest bit of run-time data for COV_ITEM_TOGGLE is used to track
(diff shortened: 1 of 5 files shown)


src/rt/cover.c
```diff
@@ -45,6 +45,13 @@ enum std_ulogic {
    _DC = 0x8
 };
 
+#define INCS_I32(i32_ptr)                          \
+      do {                                         \
+         int32_t inc = *i32_ptr + 1;               \
+         if (likely(inc > *i32_ptr))               \
+            *i32_ptr = inc;                        \
```
nickg, line 52: This doesn't work with optimisation turned on as signed types have undefined behaviour on overflow unlike unsigned types which wrap. GCC and and Clang remove the saturation check completely: https://godbolt.org/z/hY7xjaK5T

```suggestion
         if (unlikely(__builtin_add_overflow(*i32_ptr, 1, i32_ptr))) \
            *i32_ptr = INT32_MAX;                                    \
```

src/cov/cov-data.c
```diff
@@ -117,6 +118,17 @@ static cover_src_t get_cover_source(cover_item_kind_t kind, object_t *obj)
       }
    }
 
+   psl_node_t p = psl_from_object(obj);
+   if (p != NULL) {
+      printf("PSL NOT NULL!!!\n");
```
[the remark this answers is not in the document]
Blebowski: Srry, debugging left-over.

src/rt/cover.c
Blebowski: Thanks. I don't have that much experience with the nuances of the C standard, this is very much appreciated.

src/rt/cover.c
```diff
@@ -45,6 +45,13 @@ enum std_ulogic {
    _DC = 0x8
 };
 
+#define INCS_I32(i32_ptr)                                               \
+      do {                                                              \
+         int32_t inc = *i32_ptr + 1;                                    \
```
nickg, line 50:
```suggestion
```

nickg (approved):

PR #939, opened 2024-08-09
Coverage folding

Blebowski:
This PR implements coverage folding.

Coverage folding is a special type of "merging" where coverage from
equal designs placed at different places in hierarchy are merged.

This allows to merge coverage data from unit-tests into a hierarchy coverage
from a chip-top or IP-top coverage.

Each such behavior needs to be specified explicitly. This is to avoid
confusion / aliasing issues where implicit item naming (e.g. `_S0`)
may be conflicting. For this reason, only matching items / scopes
are currently merged. If part of the hierarchy below the source / destination
is distinct, such item is dropped. This is opposite to behavior with
regular merging where such item is added to the "destination" and
merging effectively creates union.

Later (in another MR), I would maybe add option to do folding
also like a "union". But since this would allow completely distinct
hierarchies to be folded (and placed into target covdb), it would
be mess... Effectively, one could completely "copy" a covdb
sub-tree, to another place in hierarchy (without the items actually
being there).

Also, later I would like to track "entity name" in the scope,
and check that the source / destination hierarchies actually
do instantiate the same entity !

This is for another MR though.


src/cov/cov-exclude.c
```diff
@@ -31,102 +31,344 @@ struct _cover_exclude_ctx {
...
+             "    Current:  %s\n"
+             "    Expected: %s\n\n",
+             istr(curr_scp->hier), istr(cmd->source));
+#endif
+
+      if (curr_scp->hier == cmd->source) {
+         cmd->found_source = true;
+         notef("Folding coverage scopes:");
+         notef("        Target - %s", istr(tgt_scope->hier));
+         notef("        Source - %s", istr(curr_scp->hier));
```
nickg, line 315: I don't think it's necessary to be this verbose. Alternatively only enable it when `DEBUG` is defined:

```suggestion
#ifdef DEBUG
         diag_t *d = diag_new(NULL, DIAG_DEBUG);
         diag_printf(d, "folding coverage scopes:");
         diag_hint(d, NULL, "        Target - %s", istr(tgt_scope->hier));
         diag_hint(d, NULL, "        Source - %s", istr(curr_scp->hier));
         diag_emit(d);
#endif
```

nvc.1
```diff
@@ -925,6 +925,75 @@ exclude WORK.TOP.SUB_BLOCK_INST.PORT_NAME.*
 # is one of the enum values used to code the FSM.
 exclude WORK.TOP.CONTROLLER.CURR_STATE.BIN_STATE.ST_ERROR
 .Ed
+.Ss Coverage folding
+NVC supports merging code coverage of single entity / module
+instantiated on multiple places in hierarchy. Such type of
+coverage merging is called coverage folding. Coverage folding
+is usefull when simulating a complex DUT. If a sub-block of a
```
nickg, line 932:
```suggestion
is useful when simulating a complex DUT. If a sub-block of a
```

nickg (changes requested):

src/cov/cov-exclude.c
```diff
@@ -30,101 +30,323 @@ struct _cover_exclude_ctx {
...
+                   "    tgt->data:           %d\n",
+                   tgt_prefix_len,
+                   src_prefix_len,
+                   istr(tgt->hier),
+                   istr(src->hier),
+                   istr(tgt_suffix_hier),
+                   istr(src_suffix_hier),
+                   src->data,
+                   tgt->data);
+#endif
```
nickg, line 233: Can you get rid of these other COVER_DEBUG_FOLD blocks too please.

Blebowski: I did. But I have the feeling I will need them in future. I can imagine the whole folding with the double
traversal of the scope hierarchy can have corner-cases that may have bugs...

nickg: It's not much effort to add them back in if/when needed.

nickg (approved):

PR #947, opened 2024-08-19
Add RPM-based dependencies to README for build-from-source

blizzardfinnegan:
This should also assist in auto-building `rpm` distribution files for RPM based package managers such as `dnf`, `yum`, `zypper`, and `yast`, as used by Fedora, CentOS, and OpenSUSE.


nickg (approved):
Thanks

PR #967, opened 2024-09-08
Use pkg-config to detect zlib

filmil:
The old code used to detect zlib by trying to compile a program. All well and good, until you want to use a zlib that is not in a standardized location.

This change allows providing zlib via a pkgconfig path.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -19,6 +19,7 @@ AM_CFLAGS += \
 	$(check_CFLAGS) \
 	$(libdw_CFLAGS) \
 	$(libffi_CFLAGS) \
+	$(zlib_CFLAGS) \
 	$(capstone_CFLAGS) \
 	$(LLVM_CFLAGS) \
 	$(libzstd_CFLAGS) \
--- a/configure.ac
+++ b/configure.ac
@@ -178,7 +178,7 @@ AC_SEARCH_LIBS([pow], [m], [], [
   AC_MSG_ERROR([unable to find the pow() function])
 ])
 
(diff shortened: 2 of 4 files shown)


filmil:
To explain a bit, I was recently trying to produce a hermetic build of nvc, and had to provide custom location flags for `zlib`. But I already had to pass special CFLAGS and LDFLAGS, so just adding to those would mess up the rest of the compilation. It's a bit of a long story, but having `zlib` be detected by pkg-config helps a lot.

filmil:
I was able to work around it by making a local patch prior to compilation. But it seems like something you'd want contributed upstream.

Makefile.am
```diff
@@ -13,12 +13,14 @@ AM_LFLAGS   = -o lex.yy.c
 
 AM_CC       = $(PTHREAD_CC)
 AM_CFLAGS   += $(PTHREAD_CFLAGS)
-AM_LDFLAGS  += $(PTHREAD_LIBS)
+AM_LDFLAGS  += $(PTHREAD_LIBS) \
+			   $(zlib_LIBS)
```
nickg, line 17: Shouldn't need to add it here, but it does need to be added to `bin_fstdump_LDADD` in `test/Makefile.am`.

nickg (commented):
Not sure why I did the check like that. Perhaps zlib didn't ship a `.pc` file at the time. Anyway this looks fine with the change suggested above to fix the Windows build.

Makefile.am
filmil: Here goes, though I will have to check what consequences it has on platforms other than Linux...

nickg (approved):

PR #989, opened 2024-09-26
src(cover): add sort to hierarchy summary tables in HTML report.

Blebowski:
A dummy sort-on-clicking-the-column-header for code coverage report tables with hierarchy summary.

I have scraped the sort function from web and modified slightly, so this is likely a sub-optimal way of doing
it, but it seems sufficient.


src/cov/cov-report.c
```diff
@@ -1029,6 +1032,73 @@ static void cover_print_hierarchy_guts(FILE *f, cover_report_ctx_t *ctx)
               "   function GetExclude(excludeCmd) {\n"
               "      navigator.clipboard.writeText(excludeCmd);\n"
               "   }\n"
+              "   function sortTable(n, tableId, is_number) {\n"
```
nickg, line 1035: This code is quite ugly and bubble sort is the worst sorting algorithm. I wrote the following based on the SO answer here https://stackoverflow.com/a/49041392

```js
function sortTable(n, tableId) {
  const table = document.getElementById(tableId);
  const rows = Array.from(table.querySelectorAll('tr:nth-child(n+2)'));

  const getCellValue = (tr) => {
    const td = tr.getElementsByTagName('td')[n];
    return td.innerText || td.textContent;
  };

  rows.sort((a, b) => {
    const left = getCellValue(this.ascending ? a : b);
    const right = getCellValue(this.ascending ? b : a);

    if (left !== '' && right !== '' && !isNaN(left) && !isNaN(right)) {
      return left - right;
    } else {
      return left.localeCompare(right);
    }
  });

  rows.forEach(tr => table.appendChild(tr));

  this.ascending = !this.ascending;
}
```

Blebowski: You are right, it is messy.

The alternative does not swap the integer percentages correctly. It treats `100 % (15/15)` as a string,
so it is the "lowest" first character. I had the `is_number` parameter for that purpose.

I updated the function to handle both string (sort by hierarchy name alphabetically) and int
(sort by box content split by `%` delimiter, sorts by number of percents reached).

src/cov/cov-report.c
```diff
@@ -1029,6 +1032,29 @@ static void cover_print_hierarchy_guts(FILE *f, cover_report_ctx_t *ctx)
               "   function GetExclude(excludeCmd) {\n"
               "      navigator.clipboard.writeText(excludeCmd);\n"
               "   }\n"
+              "   function getCellValue (tr, n) {\n"
+              "      v = tr.getElementsByTagName(\"TD\")[n];\n"
```
nickg, line 1036:
```suggestion
              "      let v = tr.getElementsByTagName(\"TD\")[n];\n"
```

Otherwise it creates a global variable.

Blebowski: Thx, I am absolute Javascript rookie. The code in `cov-report` is literally the only Javascript I ever "wrote" :D

nickg (approved):

PR #1005, opened 2024-09-30
PSL fix is unknown crash

Blebowski:


nickg:
This doesn't really solve the underlying issue as you can still get the same crash with a user-defined function:

```vhdl
entity e is
end entity;

architecture a of e is
    signal x, y : bit;

    function f (v : bit_vector) return bit is
    begin
        return v(v'left);
    end function;
begin
    assert always x -> f(y);
end architecture;
```

The problem is `sem_check` doesn't get called on the PSL HDL expressions. See bb37eb38f9.

PR #1006, opened 2024-10-02
Cover per file report

Blebowski:
Adds `--per-source-file` switch to generate per-file reports rather than hierarchical.

The feature resembles GCOV coverage reports and original NVC code coverage implementation.

Finishes last TODO from #823 .

The reshuffling to per-file is done just before the report generation. All hierarchy scopes
are traversed, and items are re-assigned to per-file report context. Consecutive items are
not corrupted since they are always added in a single scope. If an instance is used multiple
times, all its coverage data are merged (same as-if from two simulations). Merging is done
based on matching `kind`, `loc` and `flags`.

Currently this has a bug / feature that if an item is excluded, and its entity is instantiated twice,
then the item is added twice in the generated report (since excluded item is tracked in `flags`).
Maybe it is better this-way, I think it does not matter.
If someone uses exclusions, they are likely going to create hierarchy report.
If not, then pragma can be used not to emit coverage items in such source file.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,7 +4,7 @@
 - `--load` is now a global option and should be placed before the `-r`
   command.  This allows VHPI foreign subprograms to be called during
   elaboration (#988).
-- `--per-source-file` switch generates code coverage report with
+- `--per-file` switch generates code coverage report with
   code coverage shown for source files instead of hierarchies (#823).
 
 ## Version 1.14.0 - 2024-09-22
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 - `--load` is now a global option and should be placed before the `-r`
   command.  This allows VHPI foreign subprograms to be called during
   elaboration (#988).
+- `--per-source-file` switch generates code coverage report with
+  code coverage shown for source files instead of hierarchies (#823).
 
 ## Version 1.14.0 - 2024-09-22
 - Waiting on implicit `'stable` and `'quiet` signals now works
--- a/src/cov/cov-report.c
+++ b/src/cov/cov-report.c
@@ -1403,7 +1403,7 @@ static void cover_print_file_nav_tree(FILE *f, cover_rpt_file_ctx_t *ctx_list,
    for (int i = 0; i < n_ctxs; i++) {
       cover_rpt_file_ctx_t *ctx = ctx_list + i;
       // TODO: Handle paths hierarchically instead of strip!
-      const char *file_name = basename(ctx->file->name);
+      const char *file_name = basename((char *)ctx->file->name);
       fprintf(f, "<p style=\"margin-left: %dpx\"><a href=%s.html>%s</a></p>\n",
                   10, file_name, file_name);
    }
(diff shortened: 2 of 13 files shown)


Blebowski:
The `cover24` test is really dummy just to show the feature.

I have tested it with NEORV32 dummy test from `vhdl-projects`, it seems to work and data match the hierarchy report
(nothing got lost in the translation).

Maybe a good TODO would be averaging at the end, and creating only single "stats" instead of per-file stats.
Per-file stats could be then printed with `-V`.

NEWS.md
```diff
@@ -4,6 +4,8 @@
 - `--load` is now a global option and should be placed before the `-r`
   command.  This allows VHPI foreign subprograms to be called during
   elaboration (#988).
+- `--per-source-file` switch generates code coverage report with
```
nickg, line 7: Just `--per-file` is better.

src/diag.c
```diff
@@ -187,6 +187,26 @@ bool loc_invalid_p(const loc_t *loc)
       || loc->file_ref == FILE_INVALID;
 }
 
+bool locs_equal(const loc_t *lhs, const loc_t *rhs)
```
nickg, line 190: `loc_eq` already does this.

src/cov/cov-report.c
```diff
@@ -1328,10 +1355,244 @@ static void cover_report_children(cover_report_ctx_t *ctx,
...
+      }
+
+      // Print per-file report
+
+      // TODO: Handle escaped identifiers in hierarchy path!
+      // TODO: Handle paths hierarchically instead of strip!
+      char *file_name LOCAL = xasprintf("%s/%s.html", subdir, basename(ctx->file->name));
+      FILE *f = fopen(file_name, "w");
+      if (f == NULL)
+         fatal("failed to open report file: %s\n", file_name);
```
nickg, line 1559:
```suggestion
         fatal_error("failed to open report file: %s", file_name);
```

test/regress/cover24.sh
```diff
@@ -0,0 +1,14 @@
+set -xe
```
nickg, line 1: If this test doesn't actually verify anything then there's no point having it run in the regression tests. It's fine to just verify once manually that the bug is fixed.

src/diag.c
Blebowski: Sorry, I should have checked. It was stupid mistake.

test/regress/cover24.sh
Blebowski: Actually it does. With the first implementation, I had difference between numbers in NEORV32 hiearchy/file report.

The issue was, that if a cover scope had invalid loc, it did not process nested locs. Then, cover items
for else statement in `when / else` construct were not re-shuffled to per-file lists correctly. This was since
`T_COND_STMT` parent of `T_SIGNAL_ASSIGN` had no loc (which is correct since there is no condition).

So, the test-case above is MWE for this case. Before the fix, it showed one stmt cover bin less.
Its just the implementation that mixed together the original implementation and bugfix.

Blebowski: plus, if I leave the test out, the coverage will drop since there would be no test in regression that does
the `per-file` report.

nickg (approved):

PR #1032, opened 2024-10-20
OSVVM functional coverage integration

Blebowski:
Integrates OSVVM functional coverage to NVC functional coverage.
Adds default name for coverage bin if no name is set. This is needed for `CovPType`.
Adds support for changing name on the fly, since OSVVM API has such function,
and user is not forced to first create with name, and only then create cover bins.

Tracks `atleast` as threshold that needs to be reached for bin to be considered as covered.
Range for each bin is tracked. A global `threshold` option is added. Reporting is debugged:

Supports regular and cross bins. E.g.:
```
Cov1 <= NewID("DUMMY NAME") ;
Cov2 := NewID("DUMMY") ;
wait for 0 ns ; -- Update Cov1
AddCross(Cov1, GenBin(0,10000,3), GenBin(0,2), GenBin(0,2));
AddBins(Cov2, "Bin 10 to 12", GenBin(27,10,12,1));
AddBins(Cov2, "Bin 13 to 15", GenBin(13,15));
```
image
image

Needs OSVVM:
[OSVVM NVC integration](https://github.com/OSVVM/OSVVM/pull/96/files)

I guess this will wait until this is merged in OSVVM and new release is made.

### DIFF
--- a/lib/nvc/cover_pkg-body.vhd
+++ b/lib/nvc/cover_pkg-body.vhd
@@ -26,7 +26,6 @@ package body cover_pkg is
                               item      : out t_item_handle;
                               name      : in string;
                               atleast   : in integer;
-                              n_ranges  : in integer;
                               ranges    : in t_item_range_array) is
     begin
         -- Foreign subprogram
--- a/lib/nvc/cover_pkg-body.vhd
+++ b/lib/nvc/cover_pkg-body.vhd
@@ -25,7 +25,7 @@ package body cover_pkg is
     procedure add_cover_item (scope     : inout t_scope_handle;
                               item      : out t_item_handle;
                               name      : in string;
-                              atleast   : in integer;
+                              atleast   : in positive;
                               ranges    : in t_item_range_array) is
     begin
         -- Foreign subprogram
--- a/lib/nvc/cover_pkg-body.vhd
+++ b/lib/nvc/cover_pkg-body.vhd
@@ -26,7 +26,6 @@ package body cover_pkg is
                               item      : out t_item_handle;
                               name      : in string;
                               atleast   : in natural;
-                              exclude   : in boolean;
                               ranges    : in t_item_range_array) is
     begin
         -- Foreign subprogram
--- a/lib/nvc/cover_pkg-body.vhd
+++ b/lib/nvc/cover_pkg-body.vhd
@@ -22,15 +22,26 @@ package body cover_pkg is
         -- Foreign subprogram
     end procedure;
 
-    procedure add_cover_item (scope : inout t_scope_handle;
-                              item  : out t_item_handle;
-                              name  : in string) is
+    procedure add_cover_item (scope     : inout t_scope_handle;
+                              item      : out t_item_handle;
+                              name      : in string;
+                              source    : in integer;
+                              atleast   : in integer;
+                              flags     : in integer;
+                              n_ranges  : in integer;
+                              ranges    : in t_item_range_array) is
     begin
         -- Foreign subprogram
     end procedure;
 
     procedure increment_cover_item (scope : inout t_scope_handle;
                                     item  : in t_item_handle) is
+    begin
+        -- Foreign subprogram
+    end procedure;
+
+    procedure set_cover_scope_name(scope : inout t_scope_handle;
+                                   name  : in string) is
     begin
         -- Foreign subprogram
     end procedure;
--- a/lib/nvc/cover_pkg-body.vhd
+++ b/lib/nvc/cover_pkg-body.vhd
@@ -22,26 +22,26 @@ package body cover_pkg is
         -- Foreign subprogram
     end procedure;
 
-    procedure add_cover_item (scope     : inout t_scope_handle;
-                              item      : out t_item_handle;
-                              name      : in string;
-                              source    : in integer;
-                              atleast   : in integer;
-                              flags     : in integer;
-                              n_ranges  : in integer;
-                              ranges    : in t_item_range_array) is
+    procedure add_cover_item (variable scope     : inout t_scope_handle;
+                                       item      : out t_item_handle;
+                                       name      : in string;
+                                       source    : in integer;
+                                       atleast   : in integer;
+                                       flags     : in integer;
+                                       n_ranges  : in integer;
+                                       ranges    : in t_item_range_array) is
     begin
         -- Foreign subprogram
     end procedure;
 
-    procedure increment_cover_item (scope : inout t_scope_handle;
-                                    item  : in t_item_handle) is
+    procedure increment_cover_item (variable scope : inout t_scope_handle;
+                                             item  : in t_item_handle) is
     begin
         -- Foreign subprogram
     end procedure;
 
-    procedure set_cover_scope_name(scope : inout t_scope_handle;
-                                   name  : in string) is
+    procedure set_cover_scope_name(variable scope : inout t_scope_handle;
+                                            name  : in string) is
     begin
         -- Foreign subprogram
     end procedure;
--- a/lib/nvc/cover_pkg-body.vhd
+++ b/lib/nvc/cover_pkg-body.vhd
@@ -22,26 +22,24 @@ package body cover_pkg is
         -- Foreign subprogram
     end procedure;
 
-    procedure add_cover_item (variable scope     : inout t_scope_handle;
-                                       item      : out t_item_handle;
-                                       name      : in string;
-                                       source    : in integer;
-                                       atleast   : in integer;
-                                       flags     : in integer;
-                                       n_ranges  : in integer;
-                                       ranges    : in t_item_range_array) is
+    procedure add_cover_item (scope     : inout t_scope_handle;
+                              item      : out t_item_handle;
+                              name      : in string;
+                              atleast   : in integer;
+                              n_ranges  : in integer;
+                              ranges    : in t_item_range_array) is
     begin
         -- Foreign subprogram
     end procedure;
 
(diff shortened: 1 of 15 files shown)


nvc.1
```diff
@@ -667,14 +667,30 @@ a coverage bin for each state of an FSM. NVC considers internal signals of
...
+OSVVM functional coverage bin. A call to
+.Ql AddBins
+or
+.Ql AddCross
+OSVVM functions creates such NVC functional coverage bins.
+A threshold for OSVVM coverage bin to be reported as covered is given by
+.Ql AtLeast
+argument of
+.Ql AddBins
+OSVVM call.
```
nickg, line 689: Do not mention OSVVM here. As I said before this should be a generic API and not something specific to OSVVM.

nickg (changes requested):

lib/nvc/cover_pkg-body.vhd
```diff
@@ -22,9 +22,12 @@ package body cover_pkg is
...
 
-    procedure add_cover_item (scope : inout t_scope_handle;
-                              item  : out t_item_handle;
-                              name  : in string) is
+    procedure add_cover_item (scope     : inout t_scope_handle;
+                              item      : out t_item_handle;
+                              name      : in string;
+                              atleast   : in integer;
+                              n_ranges  : in integer;
+                              ranges    : in t_item_range_array) is
```
nickg, line 30:
```suggestion
                              ranges    : in t_item_range_array) is
```

You don't need the `n_ranges` argument, see the other suggested change below.

nvc.1
Blebowski: I have left just a note that user gets a hint the integration is supported.

src/rt/cover.c
```diff
@@ -388,7 +425,26 @@ void _nvc_add_cover_item(jit_scalar_t *args)
...
+   // Name remembered at the time of cover point creation in its scope
+   item->func_name = s->block_name;
+
+   item->source = COV_SRC_USER_COVER;
+   item->atleast = args[7].integer;
+   item->flags = COV_FLAG_USER_DEFINED;
+   item->n_ranges = ffi_array_length(args[8].integer);
+   item->ranges = xcalloc_array(item->n_ranges, sizeof(cover_range_t));
+
+   int32_t *ptr = (int32_t *)args[9].pointer;
```
nickg, line 437:
```suggestion
   item->n_ranges = ffi_array_length(args[10].integer);
   item->ranges = xcalloc_array(item->n_ranges, sizeof(cover_range_t));

   int32_t *ptr = (int32_t *)args[8].pointer;
```

src/rt/cover.c
```diff
@@ -342,6 +350,35 @@ void _nvc_create_cover_scope(jit_scalar_t *args)
...
+   // Rename the scope
+   LOCAL_TEXT_BUF tb = tb_new();
+   sanitise_name(tb, name_bytes, name_len);
+   ident_t name_id = ident_new(tb_get(tb));
+
+   if (s->items.count > 0)
+      jit_msg(NULL, DIAG_FATAL, "cover point name can't be set to '%s' "
+                                "since the cover point already contains bins. "
+                                "Set the name before adding any bins",
+                                istr(name_id));
```
nickg, line 375:
```suggestion
      jit_msg(NULL, DIAG_FATAL, "cannot change name of cover scope after "
              "items are created");
```

src/rt/cover.c
```diff
@@ -342,6 +350,45 @@ void _nvc_create_cover_scope(jit_scalar_t *args)
 
    ident_t name = ident_new(tb_get(tb));
    *ptr = cover_create_scope(data, parent, inst->where, name);
+
+   // Remember raw name for reporting
+   LOCAL_TEXT_BUF tb_raw = tb_new();
+   for (int i = 0; i < name_len; i++)
+      tb_append(tb_raw, name_bytes[i]);
+   if (name_len > 0)
+      (*ptr)->block_name = ident_new(tb_get(tb_raw));
```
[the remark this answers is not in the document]
Blebowski: Due to escape being handled there is one issue pending.
If user clicks on "Get Exclude Command", escaped name will be returned, and
thus can't be placed into exclude file.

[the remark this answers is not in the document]
Blebowski: But the issue is there regardless of this change

src/cov/cov-api.h
```diff
@@ -84,9 +84,15 @@ typedef enum {
    COV_SRC_CONDITION,
    COV_SRC_STATEMENT,
    COV_SRC_PSL_COVER,
+   COV_SRC_USER_COVER,
    COV_SRC_UNKNOWN,
 } cover_src_t;
 
+typedef struct {
+   int64_t  min;
+   int64_t  max;
```
nickg, line 93:
```suggestion
   uint32_t  min;
   uint32_t  max;
```

lib/nvc/cover_pkg.vhd
```diff
@@ -27,21 +27,40 @@ package cover_pkg is
...
+                                            name  : in string);
 
-    procedure increment_cover_item (scope : inout t_scope_handle;
-                                    item  : in t_item_handle);
+    procedure add_cover_item (variable scope     : inout t_scope_handle;
+                                       item      : out t_item_handle;
+                                       name      : in string;
+                                       source    : in integer;
+                                       atleast   : in integer;
+                                       flags     : in integer;
```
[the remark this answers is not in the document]
Blebowski: I have hidden those and set them inside the API with the "User defined" meanings.

lib/nvc/cover_pkg.vhd
```diff
@@ -27,21 +27,40 @@ package cover_pkg is
 
     type t_item_handle is range -1 to 2147483647;
 
+    type t_item_range is record
+        min: integer;
+        max: integer;
+    end record;
```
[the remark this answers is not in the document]
Blebowski: Thanks.

lib/nvc/cover_pkg-body.vhd
Blebowski: I see, thanks. Later on I got it based on `ffi_uarray_t`, but forgot to adapt it.

Would you mind documenting the FFI calling conventions somewhere ?
Plain comment in header file would be enough.

lib/nvc/cover_pkg-body.vhd
```diff
@@ -22,9 +22,11 @@ package body cover_pkg is
         -- Foreign subprogram
     end procedure;
 
-    procedure add_cover_item (scope : inout t_scope_handle;
-                              item  : out t_item_handle;
-                              name  : in string) is
+    procedure add_cover_item (scope     : inout t_scope_handle;
+                              item      : out t_item_handle;
+                              name      : in string;
+                              atleast   : in integer;
```
nickg, line 28: Should this be `positive`? I'm not sure it makes sense for it to be zero or negative.

Blebowski: It is as you say. However OSVVM passes it as `integer` in `VendorCovBinAdd`. So I would
maybe keep it the same type ?

nickg: You can still make it `positive` here as that's a subtype of `integer`.

Blebowski:
Please hold with merging, I would like to still add connection between Ignore Bin created via `action` and Cov item exclude flag.

Blebowski:
I added the support for illegal / ignore bins. With these the `atleast` is passed as 0, so I changed its type to `natural`.

nickg:
> Please hold with merging, I would like to still add connection between Ignore Bin created via `action` and Cov item exclude flag.

How about adding

```vhdl
type t_action is (COUNT, IGNORE, EXCLUDE);
```

and passing that instead of a `boolean`.

Blebowski:
What is the point of passing the action through the simulator boundary ?

AFAIK, ignore bins is something what should not be "taken into account",
it should not make your coverage numbers worse. So that translates to
NVCs exclude.

Illegal bins are the same thing, with optional error / fatal when such
value is encountered. I would exclude these too. If not, then reaching 100 %
coverage on a TB with illegal bins, would need getting the DUT / TB into
state that we don't want to get into intentionally.

So I don't see a point of passing this to NVC as long as we don't have
separate notion of "illegal" cover bin in the reports. I don't think that is
needed since for the reporting / measuring purposes such category does
not differ from excluded bin.

Blebowski:
Plus, the constant values provided by OSVVM (1, 0, -1) do not map to how NVC translates enum (0,1,2).
This could be done only on the C side, not in the `cover_pkg`.
So there would need to be a bit of glue logic anyway.

nickg:
Since we're using `natural` for `atleast` now can't we just repurpose zero to mean ignored? Then we don't need another argument.

Blebowski:
We could.

Would someone want to add regular bin with `atleast` equal to 0 ? It would become
automatically covered... Probably not, as it makes no sense. 

Still OSVVM allows that, and even passing full integer. I am unsure about restricting
the range of `atleast` since if there is such code, it may work in Aldec, but will report
range check error in NVC. 

I have tendency not to restrict the NVCs API more that what OSVVM allows to its
user. But maybe its wrong, I don't know.

Keeping the `exclude` argument leaves door for open for future other ways of user
exclusion. But it is the kind of thinking that goes against the YAGNI principle.

nickg:
At least zero means you don't care what the value is, so it's effectively ignored.

Blebowski:
I have saturated the `atleast` in the wrapper code. Any invalid scenario will then run and will be excluded in the report.

nickg (approved):

PR #1034, opened 2024-10-21
Warn on missing wait

NikLeberg:
Hi there!

I did (as sadly common) something stupid and my testbench did ... nothing. I recompiled with questa which gave me an nice warning that one of the processes without sensitivitiy list did not contain any wait statements. Aha! - Found the culprit.

So I thought maybe nvc would profit from a similar warning.

Currenty, this breaks some regression checks as they also trigger this warning. Should this be put behind a flag?
Let me know of any comments or nitpicks. I'm happy to change the pr in any way required.

Cheers.

### DIFF
--- a/src/sem.c
+++ b/src/sem.c
@@ -2245,10 +2245,33 @@ static void sem_check_static_elab(tree_t t)
    }
 }
 
+static void sem_check_missing_wait(tree_t t, nametab_t *tab)
+{
+   if (!opt_get_int(OPT_MISSING_WAIT))
+      return;
+
+   const int ntriggers = tree_triggers(t);
+   if (ntriggers > 0)
+      return;
+
+   const tree_flags_t flags = tree_flags(t);
+   if (flags & TREE_F_HAS_WAIT)
+      return;
+
+   diag_t *d = diag_new(DIAG_WARN, tree_loc(t));
+   diag_printf(d, "potential infinite loop in process");
+   if (!(flags & TREE_F_SYNTHETIC_NAME))
+      diag_printf(d, " %s", istr(tree_ident(t)));
+   diag_printf(d, " with no sensitivity list and no wait statements.");
+   diag_emit(d);
+}
+
 static bool sem_check_process(tree_t t, nametab_t *tab)
 {
    if (!sem_check_sensitivity(t, tab))
       return false;
+
+   sem_check_missing_wait(t, tab);
 
    return true;
 }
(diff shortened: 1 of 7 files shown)


nickg:
This is a good idea but at the moment it doesn't work for something like:

```vhdl
process is
begin
  if now > 100 ns then
    std.env.finish;
  else
    wait for 1 ns;
  end if;
end process;
```

It would be better to reuse the `TREE_F_HAS_WAIT` flag on processes as well as procedures. You can add a `tree_set_flag(proc, TREE_F_HAS_WAIT);` at line 3549 to get that. And also at the end of `sem_check_pcall`:

```c
   if (!never_waits) {
      // Procedure may wait, suppress infinite loop warning
      tree_t proc = find_enclosing(tab, S_PROCESS);
      if (proc != NULL)
         tree_set_flag(proc, TREE_F_HAS_WAIT);
   }
```

> Currenty, this breaks some regression checks as they also trigger this warning. Should this be put behind a flag?

We can add an internal option for that which is turned off in unit tests. Grep for `OPT_MISSING_BODY` to see an example.

NikLeberg:
I initially saw the `TREE_F_HAS_WAIT` and thought this would be super easy as everything seemed to be already in place. But yeah, it was not set for processes. Thanks for pointing out where to extend it to also set it for processes. See updated commits.

Locally the checks pass, and the new test for the warning is also passing. Let me know if you come up with other test cases that we might want to add.

src/sem.c
```diff
@@ -2245,11 +2245,33 @@ static void sem_check_static_elab(tree_t t)
    }
 }
 
+static bool sem_check_missing_wait(tree_t t, nametab_t *tab) {
```
nickg, line 2248:
```suggestion
static void sem_check_missing_wait(tree_t t, nametab_t *tab) 
{
```

The `return false` is used to indicate an error and sometimes skips subsequent checks, but we don't want that for just a warning so make it `void` instead. Also the `{` goes on a new line for functions.

src/sem.c
```diff
@@ -2245,11 +2245,33 @@ static void sem_check_static_elab(tree_t t)
...
+
+   return true;
+}
+
 static bool sem_check_process(tree_t t, nametab_t *tab)
 {
    if (!sem_check_sensitivity(t, tab))
       return false;
 
+   if (opt_get_int(OPT_MISSING_WAIT) && !sem_check_missing_wait(t, tab))
```
nickg, line 2272: I'd put the `OPT_MISSING_WAIT` check inside `sem_check_missing_wait`. Seems cleaner that way.

test/test_bounds.c
```diff
@@ -31,6 +31,7 @@
 START_TEST(test_bounds)
 {
    opt_set_int(OPT_MISSING_BODY, 0);
+   opt_set_int(OPT_MISSING_WAIT, 0);
```
nickg, line 34: As there's so many of these it would be better to do it once in `main()` in `test/unit_test.c` and then turn it on for the one test that needs it.

NikLeberg: That's a nicer way to do it for sure. 😅

nickg (approved):

Issue #912, opened 2024-07-07
PSL sequence implication parsing error


Blebowski:
I run into trouble analysis PSL assertions that contain sequence implications.

E.g.:
```
library ieee;
use ieee.std_logic_1164.all;

entity psl_sequence_implication is
end entity;

architecture tb of psl_sequence_implication is

    signal clk : std_logic;

    signal a,b : std_logic;

begin

    -- psl default clock is rising_edge(clk);

    -- psl asrt_1 : assert always
    --  {a = '1'} |=> {b = '1'};

end architecture;
```

I get:
```
nvc -a --psl psl_sequence_implication.vhd 
** Error: unexpected | while parsing assert directive, expecting one of @ or ;
    > psl_sequence_implication.vhd:18
    |
 18 |     --  {a = '1'} |=> {b = '1'};
    |                   ^ this token was unexpected
```

The code analyses OK in GHDL and Aldec.

PR #1039, opened 2024-10-25
Psl seq implication

Blebowski:
Adds simple support for PSL sequence implication. FSMs for LHS and RHS are
created separately and merged. A default transition is added from each state of
LHS FSM into `accept` state of the RHS FSM.

I have added `P_SEQ_IMPLICATION` instead of encoding it into the `P_IMPLICATION`,
it seemed better to separate it.


Blebowski:
https://github.com/nickg/nvc/issues/912

nickg:
Thanks, but this isn't how I intended to do the FSM generation. I implemented this in b5b15487 with some bits from this patch.

Blebowski:
No problem, thanks for the implementation.

Would you mind incorporating changes from:
[d68fe2786461df2bc9f6dde7fd253bafeab22509](https://github.com/nickg/nvc/pull/1039/commits/d68fe2786461df2bc9f6dde7fd253bafeab22509)

PR #1041, opened 2024-10-27
Optimize toggle coverage performance

Blebowski:
Optimizes toggle coverage performance.

On large signals (e.g. with `--include-mems`), a loop + check through each byte is
performance bottleneck. The optimization can use the fact that it is likely that not all
signals of a vector change at one time.

Instead, signal bits are split into batches of 8 bytes. An iteration through bytes of batch
is done only when a change occurred in the batch. A change detection is fast since
batch size matches word size. Further, each call to `check_fnc` is done only when
`old_val` and `new_val` differ to prevent falling through all the `if/else` statements if
nothing has changed.

For smaller sized signals the optimization is disabled since these are likely to change
all in a single callback (e.g. address / data bus, etc...).

On common simulation there is almost no visible effect (since optimization is disabled).
E.g. on:
```
library ieee;
use IEEE.std_logic_1164.all;

entity cover26 is
end entity;

architecture test of cover26 is

    type t_mem is array (32767 downto 0) of std_logic_vector(31 downto 0);

    signal mem : t_mem := (others => (others => '0'));

begin

    process
    begin
        for i in 0 to 32767 loop
            mem(i)(0) <= '1';
            wait for 1 ns;
        end loop;
        wait;
     end process;

end architecture;
```
when running with `include-mems` there is 50 % performance improvement.

### DIFF
--- a/src/rt/cover.c
+++ b/src/rt/cover.c
@@ -124,8 +124,8 @@ static inline void cover_toggle_check_0_1_u_z(uint8_t old, uint8_t new,
 #ifdef COVER_DEBUG_CALLBACK
 #define COVER_TGL_CB_MSG(signal)                                              \
    do {                                                                       \
-      printf("Time: %lu Callback on signal: %s\n",                            \
-              now, istr(tree_ident(signal->where)));                          \
+      printf("Time: %lu Callback on signal: %s with size: %d \n",             \
+              now, istr(tree_ident(signal->where)), s->shared.size);          \
    } while (0);
 
 #define COVER_TGL_SIGNAL_DETAILS(signal, size)                                \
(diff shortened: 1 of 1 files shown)


src/rt/cover.c
```diff
@@ -145,25 +145,68 @@ static inline void cover_toggle_check_0_1_u_z(uint8_t old, uint8_t new,
...
-         uint8_t new = ((uint8_t*)signal_value(s))[i];                        \
-         uint8_t old = ((uint8_t*)signal_last_value(s))[i];                   \
-         check_fnc(old, new, toggle_01, toggle_10);                           \
-         toggle_01 += 2;                                                      \
-         toggle_10 += 2;                                                      \
-      }                                                                       \
-      COVER_TGL_SIGNAL_DETAILS(s, s_size)                                     \
-   }                                                                          \
+// Set to match word size of common PC and optimize for int64_t comparison
+#define TOGGLE_BATCH_SIZE 8
```
nickg, line 149: Use `sizeof(uint64_t)` instead of this. It's shorter and a `#define` might lead someone to wrongly assume that it can be changed when the only possible value is 8.

src/rt/cover.c
```diff
@@ -145,25 +145,68 @@ static inline void cover_toggle_check_0_1_u_z(uint8_t old, uint8_t new,
...
+         int32_t *toggle_10 = toggle_01 + 1;                                        \
+         for (int i = 0; i < s_size; i++) {                                         \
+            uint8_t new = ((uint8_t*)signal_value(s))[i];                           \
+            uint8_t old = ((uint8_t*)signal_last_value(s))[i];                      \
+            check_fnc(old, new, toggle_01, toggle_10);                              \
+            toggle_01 += 2;                                                         \
+            toggle_10 += 2;                                                         \
+         }                                                                          \
+         return;                                                                    \
+      }                                                                             \
```
nickg, line 177: Why bother with this? For small sizes it falls through to the scalar loop on line 198 which does exactly the same thing. For sizes in 8..128 there's presumably some benefit to comparing eight bytes at a time, certainly more than the overhead of the extra branches.

src/rt/cover.c
```diff
@@ -145,25 +145,68 @@ static inline void cover_toggle_check_0_1_u_z(uint8_t old, uint8_t new,
...
+      /* Optimize for assumption that most bits don't change in large signals */    \
+      uint32_t batches = ((s_size - 1) / TOGGLE_BATCH_SIZE) + 1;                    \
+      for (int i = 0; i < batches; i++) {                                           \
+         bool walk = false;                                                         \
+         int batch_size = TOGGLE_BATCH_SIZE;                                        \
+         if (i < batches - 1) {                                                     \
+            const void *batch_new = signal_value(s) + i * TOGGLE_BATCH_SIZE;        \
+            const void *batch_old = signal_last_value(s) + i * TOGGLE_BATCH_SIZE;   \
+            if (*((int64_t*)batch_new) != *((int64_t*)batch_old))                   \
```
nickg, line 186:
```suggestion
            if (unaligned_load(batch_new, uint64_t) != unaligned_load(batch_old, uint64_t))                   \
```

I think the current value pointer is always 64-bit aligned at the moment but the last value pointer won't necessarily be.

src/rt/cover.c
Blebowski: The idea was that in a most designs out there, there are either 32 or 64 bit SoC buses. Such buses likely change most bits of the vector during a single event (e.g. a new instruction fetched from memory). Then it is redundant to do the word compare first, and then walk each byte on compare as we well almost always need to do the walk.

Blebowski: And I chose 128 with some reserve.

I think a better optimization could be based not on signal size, but based on signal being multidimensional or not. The multidimensional signals are likely to be memories where only single row changes at a given time.

Blebowski: I would keep the current format, but it is up to you. I can remove the threshold completely.

nickg: Have you actually measured any performance improvement by doing this? Here's the NEORV32 test bench run with `--cover=toggle` and the code in this PR:

```
  Time (mean ± σ):     7073.3 ms ±  53.7 ms    [User: 7054.7 ms, System: 17.6 ms]              
  Range (min … max):   7021.0 ms … 7142.0 ms    5 runs                              
```

And if I set `TOGGLE_OPT_TH` to zero:

```
  Time (mean ± σ):     7010.8 ms ±  59.8 ms    [User: 6992.7 ms, System: 15.8 ms]
  Range (min … max):   6964.1 ms … 7108.9 ms    5 runs
```

Which is actually slightly faster (although not statistically significant).

For small arrays the main cost seems to be the cache miss loading the counter before incrementing as it's indexed by the tag which is essentially random and so hard to prefetch. Once you've loaded the eight bytes of signal data it's already in L1 and the cost of loading it again a byte at a time is only around four cycles on modern CPUs which is pretty insignificant. 

On the other hand removing the `TOGGLE_OPT_TH` check increases the size and complexity of this macro by about 50% which is definitely not worth it.

Blebowski: > Have you actually measured any performance improvement by doing this?

No. It was last minute change trying to avoid extra work in cases where it will likely not be needed. I should have done some benchmark.

I only measured the dummy example above when I was thinking on how to optimize the code.

> For small arrays the main cost seems to be the cache miss loading the counter before incrementing as it's indexed by the tag which is essentially random and so hard to prefetch. Once you've loaded the eight bytes of signal data it's already in L1 and the cost of loading it again a byte at a time is only around four cycles on modern CPUs which is pretty insignificant.
 
Well, that makes sense, and I did not think about it that way. Doing the 64 bit compare is one instruction, so no overhead. Loading of signal data needs to be done either way, so no savings.

> On the other hand adding the TOGGLE_OPT_TH check increases the size and complexity of this macro by about 50% which is definitely not worth it.

Yeah, it is ugly, and as it has no provable benefit I will rather remove it.

Issue #1046, opened 2024-10-30
Problem building on macos, llvm-config from llvm-16 doesn't include -L for zstd


jeisch:
On Sonoma 14.7, the standard build procedure fails when running `../configure` at:
```
checking for LLVM (engine passes ipo linker native)... no
configure: error: Could not detect the LLVM libraries. Make sure that llvm-config is on your path or specified by --with-llvm.
```

From the config.log, this is actually a linker error with zstd:

```
configure:10362: checking for LLVM (engine passes ipo linker native)
[...]

PR #1047, opened 2024-10-30
Fix for https://github.com/nickg/nvc/issues/1046 on macos

jeisch:
Adding LDFLAGS for zstd from pkg-config to resolve issue https://github.com/nickg/nvc/issues/1046

### DIFF
--- a/m4/ax_llvm_c.m4
+++ b/m4/ax_llvm_c.m4
@@ -118,6 +118,15 @@ AC_DEFUN([AX_LLVM_C], [
                          [LLVM uwtable attribute takes an argument])
     fi
 
+    if test "$llvm_ver_num" -ge "160"; then
+      case $host_os in
+      darwin*)
+        # Add LDFLAGS for libzstd in a non-default location for issue #1046
+        ZSTD_LDFLAGS=`pkg-config --libs-only-L libzstd`
+        LLVM_LDFLAGS="$LLVM_LDFLAGS $ZSTD_LDFLAGS"
+      esac
+    fi
+
     LLVM_OBJ_EXT="o"
     case $host_os in
       *cygwin*|msys*|mingw32*)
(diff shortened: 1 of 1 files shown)


m4/ax_llvm_c.m4
```diff
@@ -118,6 +118,15 @@ AC_DEFUN([AX_LLVM_C], [
                          [LLVM uwtable attribute takes an argument])
     fi
 
+    if test "$llvm_ver_num" -ge "160"; then
+      case $host_os in
+      darwin*)
+        # Add LDFLAGS for libzstd in a non-default location for issue #1046
+        ZSTD_LDFLAGS=`pkg-config --libs-only-L libzstd`
```
nickg, line 125: It's probably better to use the `PKG_CHECK_MODULES` macro here. Does this work?

```
PKG_CHECK_MODULES([libzstd], [libzstd >= 1.4])
LLVM_LDFLAGS="$LLVM_LDFLAGS $libzstd_LDFLAGS"
```

nickg: Probably should be `LLVM_LIBS` instead of `LLVM_LDFLAGS`.

jeisch:
I have incorporated the comments into my most recent commit to the branch and tested successfully on my machine.

nickg:
I moved this around a bit in 41d7f9f2f40f to avoid a warning about duplicated `-lstd`. Can you check it still works for you?

jeisch:
On my system, `-lzstd` is provided by both `pkg-config --libs libzstd` (the source of $libzstd_LIBS) and by `llvm-config --system-libs` so it is still duplicated after the changes in https://github.com/nickg/nvc/commit/41d7f9f2f40ff48307a2fbdbb464cdcaa41d6c17 

>  % pkg-config --libs libzstd
-L/opt/local/lib -lzstd
 % llvm-config --system-libs
-lm -lz -lzstd -lcurses -lxml2

nickg:
OK I think I misunderstood the problem a bit and missed that you have `--libs-only-L` in the original patch. Can you test #1056? It's basically the same as your original patch but without hard-coding `pkg-config`.

PR #1048, opened 2024-11-01
Fix crash on warning of hidden declaration. Issue #1038

NikLeberg:
The fuzzing done by @avelure produced various interesting errors.

The one I try to fix here seems to stem from a simple typo in the implementation of `warn_hidden_decl`.

I added a testcase but it produces another error `../test/test_util.c:199:F:Core:test_issue1038:0: Assertion 'parse() == NULL' failed` But I don't know exactly why. I do not fully understand the test framework though and could be doing something stupid. 

Cheers

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,6 +15,7 @@
 - Added a warning for potential infinite loops in processes without
   sensitivity and lacking any wait statements (from @NikLeberg).
 - Fixed a crash when `'last_value` is used with record types (#1043).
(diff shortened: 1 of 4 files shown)


NikLeberg:
see https://github.com/nickg/nvc/issues/1038#issuecomment-2451729115 for reference

nickg:
`parse_and_check(T_PACKAGE, T_PACK_BODY)` should fix the assertion failure.

nickg (approved):

PR #1050, opened 2024-11-02
Add NULL check before calling type_eq.

NikLeberg:
See https://github.com/nickg/nvc/issues/1038#issuecomment-2452911050.

This one happened when `type_eq` was called on a `kind` that is actually `NULL`. I added a simple check. The tests with `make check` are passing and seem to indicate that nothing else was impacted. Lets see what CI thinks of it.

I preliminary merged #1048 to build upon the same test case. I'll rebase once that is in master.

Cheers

### DIFF
--- a/test/test_parse.c
+++ b/test/test_parse.c
@@ -6919,7 +6919,7 @@ START_TEST(test_issue1038)
 
    input_from_file(TESTDIR "/parse/issue1038.vhd");
 
-   parse_and_check(T_PACKAGE, T_PACK_BODY);
+   parse_and_check(T_PACKAGE, T_PACK_BODY, T_PACKAGE);
 
    fail_unless(parse() == NULL);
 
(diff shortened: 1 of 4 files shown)


test/test_parse.c
```diff
@@ -6912,6 +6912,27 @@ START_TEST(test_issue991)
...
 
+START_TEST(test_issue1038)
+{
+   opt_set_int(OPT_RELAXED, 1);
+   set_standard(STD_08);
+
+   input_from_file(TESTDIR "/parse/issue1038.vhd");
+
+   const error_t expect[] = {
+      { 14, "aliased name is not static" },
```
nickg, line 6923: I think this is a bug too, a unit name should be a static name. Probably need to add `T_UNIT_DECL` to the list in `sem_static_name`.

nickg (commented):
Needs a rebase after I merged the other PR.

test/test_parse.c
NikLeberg: I've added an additional commit for it. Or would you like it to be an individual PR?

nickg: It's fine in this one.

nickg (approved):

PR #1051, opened 2024-11-02
Fix crash on unexpected psl builtin function in primary

NikLeberg:
See https://github.com/nickg/nvc/issues/1038#issuecomment-2452959299.

While parsing an unexpected PSL builtin function in a primary nvc does crash on an assertion. I relaxed the assertion to an if check. If the check fails, then the switch falls-through to the already existing error message.

Cheers

### DIFF
--- a/src/lexer.l
+++ b/src/lexer.l
@@ -570,7 +570,7 @@ BEFORE            ?i:before
 <PSL>{IS}                { TOKEN(tIS); }
 <PSL>"->"                { TOKEN(tIFIMPL); }
 <PSL>"<->"               { TOKEN(tIFFIMPL); }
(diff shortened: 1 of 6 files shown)


nickg:
A neater fix to this would be to define a new `tPSLNEXT` token so it can't be confused with the VHDL `tNEXT`. So in `lexer.l`:

```c
<PSL>{NEXT}              { TOKEN(tNEXT); }
```

Would be replaced with

```c
<PSL>{PSLNEXT}           { TOKEN(tPSLNEXT); }
```

nickg:
Need to add "next" to the end of the long array in `token_str` in `src/scan.c`, otherwise looks fine.

nickg (approved):

PR #1053, opened 2024-11-03
PSL SERE repetition

Blebowski:
Hi,

this is a WIP of sequence repetition, not yet ready to be merged.

I would like to discuss how to proceed with the FSM generation. Consider the example below:
```
-- psl cov_1 : cover {a;b}[*2 to 3] report "COVER HIT!";
```
The current code leads to FSM like so:
image

State 1 serves as skip condition when minimal required number of repetitions were already met.
In state 8, I would need to:
- Move to 1 via epsilon edge since two repetitions were hit
- Move to 9 since I would like to track third repetition from next cycle on.

Both of these are un-guarded. I create 8 -> 9 edge during "repeat" part, so it is not guarded since
checking of "third" repetition is by guard on epsilon 9 -> 10.

The `psl_lower_state` fires an assert, since I have two "default" edges, one `EDGE_NEXT`
and one `EDGE_EPSILON`.

I need to activate both 1 and 9 from 8, so "branch" the FSM traverse.
AFAICT, the `psl_lower_state` generates code in `if-elsif-else` manner, making sure only one transition
is activated.

I think the behavior could be:
- All `EDGE_NEXT` edges that have guard are activated if the guard is 1 (instead of the first one matching)
- If `EDGE_NEXT` has no guard, it is always activated.
- `EDGE_EPSILON` will behave as till now (activated in `if-else-if` manner, there can be only one default `EDGE_EPSILON`).

I think the branching behavior will be needed also for some of the other SERE logic (`within` or `&&`).

Maybe some of the PSL syntax will require also multiple activated `EDGE_EPSILON`. But since this is
done with `emit_jump`, it would require more complex logic (some sort of call). Anyway, this is now not
needed, only the points above.

Let me know what do you think. Btw. thanks for all the effort in the recent weeks.

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -4585,7 +4585,7 @@ static tree_t p_primary(tree_t head)
 {
    // name | literal | aggregate | function_call | qualified_expression
    //   | type_conversion | allocator | ( expression ) |
-   //  PSL: Built_In_Function_Call
+   //   | PSL: Built_In_Function_Call
 
    BEGIN("primary");
 
@@ -11781,8 +11781,29 @@ static tree_t p_psl_count(void)
 
    tree_t count = p_expression();
 
-   if (peek() == tTO)
-      count = p_range(count);
+   if (peek() == tTO) {
+      if (peek_nth(2) == tINF) {
+         consume(tTO);
+
+         tree_t r = tree_new(T_RANGE);
+         tree_set_subkind(r, RANGE_TO);
+
+         tree_t inf = tree_new(T_LITERAL);
+
+         consume(tINF);
+
+         tree_set_loc(inf, CURRENT_LOC);
+         tree_set_subkind(inf, L_INT);
+         tree_set_ival(inf, INT32_MAX);
+
+         tree_set_left(r, count);
+         tree_set_right(r, inf);
+
+         count = r;
+      }
+      else
+         count = p_range(count);
+   }
 
    solve_types(nametab, count, std_type(NULL, STD_INTEGER));
 
@@ -11800,7 +11821,7 @@ static psl_node_t p_psl_repeat_scheme(void)
 
    psl_node_t rpt = psl_new(P_REPEAT);
 
-   const token_t tok = one_of(tPLUSRPT, tTIMESRPT, tGOTORPT, tARROWRPT);
+   const token_t tok = one_of(tPLUSRPT, tTIMESRPT, tGOTORPT, tEQRPT);
    switch (tok) {
    case tPLUSRPT:
       psl_set_subkind(rpt, PSL_PLUS_REPEAT);
@@ -11819,8 +11840,8 @@ static psl_node_t p_psl_repeat_scheme(void)
       consume(tRSQUARE);
       break;
 
-   case tARROWRPT:
-      psl_set_subkind(rpt, PSL_ARROW_REPEAT);
+   case tEQRPT:
+      psl_set_subkind(rpt, PSL_EQUAL_REPEAT);
       if (peek() != tRSQUARE)
          psl_set_tree(rpt, p_psl_count());
       consume(tRSQUARE);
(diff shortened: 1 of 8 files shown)


nickg:
> 
> I think the behavior could be:
> 
>     * All `EDGE_NEXT` edges that have guard are activated if the guard is 1 (instead of the first one matching)
> 
>     * If `EDGE_NEXT` has no guard, it is always activated.
> 
>     * `EDGE_EPSILON` will behave as till now (activated in `if-else-if` manner, there can be only one default `EDGE_EPSILON`).
> 

Yeah I think something like this will work. I was planning to add FL `and` soon which also needs this.

Blebowski:
I have implemented this change. The tracking of "vanishing FSM state" is now done
in run-time code since the `psl_enter_state` does not call `vcode_return` anymore.
Thus `psl_enter_state` falls-through to the assertion checking that sequence is not met.

Implements full consecutive repetition of PSL sequences. An open-ended repetition with `inf`
is supported by FSM looping to "last but one" repetition of the sequence.

The MR is now ready for review.

src/parse.c
```diff
@@ -4357,6 +4358,17 @@ static tree_t p_literal(void)
...
+      assert(is_scanned_as_psl());
+      tree_t inf = tree_new(T_LITERAL);
+
+      consume(tINF);
+      tree_set_loc(inf, CURRENT_LOC);
+      tree_set_subkind(inf, L_INT);
+      tree_set_ival(inf, INT32_MAX);
+
+      return inf;
+
```
nickg, line 4371:
```suggestion
   case tINF:
      {
         assert(is_scanned_as_psl());
         tree_t inf = tree_new(T_LITERAL);

         consume(tINF);
         tree_set_loc(inf, CURRENT_LOC);
         tree_set_subkind(inf, L_INT);
         tree_set_ival(inf, INT32_MAX);

         return inf;
      }

```

src/psl/psl-fsm.c
```diff
@@ -215,13 +215,11 @@ static fsm_state_t *build_sere(psl_fsm_t *fsm, fsm_state_t *state, psl_node_t p)
 {
    const int nops = psl_operands(p);
 
-   fsm_state_t *initial = state;
-
-   for (int i = 0; i < nops; i++) {
-      psl_node_t rhs = psl_operand(p, i);
+   for (int j = 0; j < nops; j++) {
```
nickg, line 218: No need to change the variable name

Blebowski: It was a left-over from previous implementation approach where this loop was nested within another loop.

src/psl/psl-lower.c
```diff
@@ -114,37 +114,48 @@ static void psl_lower_state(lower_unit_t *lu, psl_fsm_t *fsm,
    if (state->initial && psl_fsm_repeating(fsm))
       psl_enter_state(state);
 
+   vcode_type_t vbool = vtype_bool();
+   vcode_reg_t vfalse = emit_const(vbool, 0);
+   vcode_reg_t vtrue = emit_const(vbool, 1);
+   vcode_var_t vanish = emit_var(vbool, vbool, ident_new("vanish"), VAR_TEMP);
```
nickg, line 120: You don't need a variable here, the guard expressions can be OR-ed together. I pushed this separately in 8806798.

src/parse.c
```diff
@@ -4357,6 +4358,19 @@ static tree_t p_literal(void)
...
+         tree_t inf = tree_new(T_LITERAL);
+
+         consume(tINF);
+         tree_set_loc(inf, CURRENT_LOC);
+         tree_set_subkind(inf, L_INT);
+         tree_set_ival(inf, INT32_MAX);
+
+         return inf;
+      }
+
```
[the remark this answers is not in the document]
Blebowski: I have moved it to `p_psl_count`.

src/psl/psl-lower.c
Blebowski: I dropped my changes to `psl_lower_state` and rebased.

nickg (approved):

PR #1061, opened 2024-11-07
Fix crash on long source lines.

NikLeberg:
Me with the fuzzing fixes again.
See https://github.com/nickg/nvc/issues/1038#issuecomment-2462371262.

The _extremely_ long line contains an error at the very end of it. This causes the `loc_t` for the diagnostic to be invalid. This then causes the later call to `diag_hints` to crash.
By simply checking if there exists any hints at all in `diag_hints`, the crash can be avoided.

But the output of the MWE is now:
```
** Error: no visible subprogram declaration for NATURAL
** Note: NATURAL called here
```
It does not crash, but the error/note are not really helpful as the line number is missing.
As I don't think that this error case would be common, I think this is acceptable. What do you think?

Cheers


NikLeberg:
Ah I see just now, it would be capable of just emitting the line number of the error. Seems as if something else is causing the `loc_t` to be off. I'll have a look.

nickg:
On 2024-11-07 16:37, NikLeberg wrote:
> Ah I see just now, it would be capable of just emitting the line
> number of the error. Seems as if something else is causing the loc_t
> to be off. I'll have a look.
> 

I think the check at the start of `get_loc` is wrong. Try the following:

```c
    if (first_line > last_line)
       return LOC_INVALID;
    else	if (first_line == last_line && first_column > last_column)
       return LOC_INVALID;
```

NikLeberg:
> I think the check at the start of `get_loc` is wrong. Try the following:

Nice one! That fixes it.

nickg:
It's still worth keeping the check in `diag_hints` in case that can be hit another way.

nickg (approved):

PR #1063, opened 2024-11-08
Cover fix long report names

Blebowski:
This MR removes limit of 256 characters in a hierarchy where coverage is reported.

The coverage reports are created hierarchically. Each hierarchy creates a nested folder.
Links to source files or other hierarchy levels are properly prefixed with `..` to point
to correct files.


src/cov/cov-report.c
```diff
@@ -308,11 +309,14 @@ static void cover_print_html_header(FILE *f)
...
 
-static void cover_print_file_name(FILE *f, const char *name)
+static void cover_print_file_name(FILE *f, int levels, const char *name)
 {
    fprintf(f, "<h2 style=\"margin-left: " MARGIN_LEFT ";\">\n");
-   fprintf(f, "   File:&nbsp; <a href=\"../../%s\">../../%s</a>\n",
-            name, name);
+   fprintf(f, "   File:&nbsp; <a href=\"");
+   for (int i = 0; i <= levels; i++)
+      fprintf(f, "..%s", DIR_SEP);
```
nickg, line 317: This won't work on Windows: the separator in a URL should always be `/`.

src/cov/cov-report.c
```diff
@@ -1622,23 +1642,25 @@ static void cover_report_per_file(FILE *top_f, cover_data_t *data, char *subdir)
...
-      FILE *f = fopen(file_name, "w");
-      if (f == NULL)
-         fatal_errno("failed to open report file: %s\n", file_name);
+      const char *src_file_name = basename((char*)ctx->file->name);
+      const char *file_rpt_path LOCAL = xasprintf("%s%s%s.html", subdir, DIR_SEP, src_file_name);
 
-      ident_t file_name_id = ident_new(basename((char*)ctx->file->name));
+      FILE *f = fopen(file_rpt_path, "w");
+      if (f == NULL)
+         fatal_errno("failed to open report file: %s\n", file_rpt_path);
```
nickg, line 1650:
```suggestion
         fatal_errno("failed to open report file: %s", file_rpt_path);
```

src/cov/cov-report.c
```diff
@@ -1696,7 +1717,7 @@ void cover_report(const char *path, cover_data_t *data, int item_limit)
    notef("Code coverage report folder: %s.", path);
    notef("%s", tb_get(tb));
 
-   char *top LOCAL = xasprintf("%s/index.html", path);
+   char *top LOCAL = xasprintf("%s%sindex.html", path, DIR_SEP);
```
nickg, line 1720: Leave it as `/`. The C library on Windows accepts both.

test/regress/testlist.txt
```diff
@@ -1068,3 +1068,4 @@ issue1044       mixed
 issue1030       wave,2008,dump-arrays
 issue1036       normal
 issue1057       normal,vhpi,2008
+cover26         shell
```
nickg, line 1071:
```suggestion
cover26         cover,gold
```

Adding "cover" here makes `run_regr` do exactly what is in the shell script.

src/cov/cov-report.c
Blebowski: I see, the docs for `fopen` do clarify it. Sorry for the HTML, I did not realize that.
What about the `make_dir`. The underlying `CreateDirectory` docs I was able to find
only mention that `/` can be used on Unix systems.

nickg:
There's still a limit of 260 characters in the full path name on Windows. Perhaps it would be better to hash the hierarchical path to generate a file name. That also avoids issues with special characters in extended identifiers.

Blebowski:
> There's still a limit of 260 characters in the full path name on Windows. Perhaps it would be better to hash the hierarchical path to generate a file name. That also avoids issues with special characters in extended identifiers.

I did not realize that. I have not done anything on Windows since 2018.

I will close the PR and create another one, as the current master is better base for such change.
The changes to find proper relative hierarchy level are redundant then.

PR #1082, opened 2024-11-28
Fix crash on invalid subtype indication.

NikLeberg:
See https://github.com/nickg/nvc/issues/1038#issuecomment-2465092738.

In `resolve_resolution` we expect `rname` to be of kind `T_REF` or `T_ELEM_RESOLUTION` otherwise an assert triggers. I simply added a check for the _kind_ and prevent calling `resolve_resolution`. If this erroreous input is now given, the parser no longer crashes but instead maneuvers completely off the line and everything afterwards may generate lots of other errors. For such a rare and _mean_ input I think its ok though. What do you think?

Cheers


nickg:
I think it would be better to adjust `resolve_resolution` instead so that it handles this case. Something like

```c
switch (tree_kind(rname)) {
case T_ELEM_RESOLUTION:
   ...
case T_REF:
   ...
default:
   error_at(tree_loc(rname), "not a valid resolution function name");
   tree_set_type(rname, type_new(T_NONE));
   break;
```

PR #1083, opened 2024-11-28
Fix crash on package definition in interface decl.

NikLeberg:
My own instance of a fuzzer (libFuzzer based, [nvc-fuzz](https://github.com/NikLeberg/nvc-fuzz)) found a crash.

If instead of the usual `package <x> is new ...` in an interface a full package definition `package <x> is ... end package;` was given, then parsing did crash.

With the added simple checks, the crash no longer happens and the parser gives the reasonable error of `unexpected end while parsing interface package declaration, expecting new`.

Note: The standard is a bit moot in the definition, but I think to understand it so that generic package in interfaces may only be defined in generic interfaces.
> An interface package declaration declares an interface package that appears in a generic clause. [VHDL2019, Chapter 6.5.5]

But NVC allows them everywhere. E.g. the following is analyzing, elaborating and running without error:
```vhd
package gen_pkg is
    generic (N : integer);
    constant c_pkg : integer := N;
end package;

entity ent is
    generic (
        package p0 is new work.gen_pkg generic map(<>);
    );
    port (
        package p1 is new work.gen_pkg generic map(<>); -- no error?
    );
end entity;

architecture arch of ent is
    constant c0 : integer := p0.c_pkg;
    constant c1 : integer := p1.c_pkg;
begin
end architecture;
```

Cheers

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -5541,14 +5541,15 @@ static void p_formal_parameter_list(tree_t decl, type_t type)
    if (nports == 0)
       return;   // Was parse error
 
-   tree_t p0 = tree_port(decl, 0);
-   type_add_param(type, tree_type(p0));
-
-   if (tree_has_value(p0))
-      tree_set_flag(decl, TREE_F_CALL_NO_ARGS);
-
-   for (int i = 1; i < nports; i++)
-      type_add_param(type, tree_type(tree_port(decl, i)));
+   for (int i = 0; i < nports; i++) {
+      tree_t p = tree_port(decl, i);
+      if (i == 0 && tree_has_value(p))
+         tree_set_flag(decl, TREE_F_CALL_NO_ARGS);
+      if (tree_has_type(p))
+         type_add_param(type, tree_type(p));
+      else
+         type_add_param(type, type_new(T_NONE));   // Will raise error later
+   }
 }
 
 static tree_t p_interface_function_specification(void)
--- a/NEWS.md
+++ b/NEWS.md
@@ -34,6 +34,7 @@
   format (#1071).
 - Added support for package external names (#1072).
 - Fixed parser hang for bit string literals (from @NikLeberg).
+- Fixed a crash on an illegal package definition inside an interface.
 - Several other minor bugs were resolved (#1038, #1057, #1067).
 
 ## Version 1.14.2 - 2024-11-23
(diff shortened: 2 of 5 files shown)


src/parse.c
```diff
@@ -5536,13 +5536,17 @@ static void p_formal_parameter_list(tree_t decl, type_t type)
       return;   // Was parse error
 
    tree_t p0 = tree_port(decl, 0);
-   type_add_param(type, tree_type(p0));
+   if (tree_has_type(p0))
+      type_add_param(type, tree_type(p0));
```
nickg, line 5540:
```suggestion
   if (class_has_type(tree_class(p0)))
      type_add_param(type, tree_type(p0));
   else
      type_add_param(type, type_new(T_NONE));   // Will raise error later
```

src/parse.c
```diff
@@ -5536,13 +5536,17 @@ static void p_formal_parameter_list(tree_t decl, type_t type)
...
 
    if (tree_has_value(p0))
       tree_set_flag(decl, TREE_F_CALL_NO_ARGS);
```
nickg, line 5543: Maybe we can merge this into the loop below to avoid duplicating the check above. E.g.

```c
if (i == 0 && tree_has_value(p))
   tree_set_flag(decl, TREE_F_CALL_NO_ARGS);
```

nickg:
> Note: The standard is a bit moot in the definition, but I think to understand it so that generic package in interfaces may only be defined in generic interfaces.

I think the relevant bit is in 6.5.6.1:

> A generic interface list consists entirely of interface constant declarations, interface type declarations, interface subprogram declarations, and interface package declarations. A port interface list consists entirely of interface signal declarations. A parameter interface list may contain interface constant declarations, interface signal declarations, interface variable declarations, interface file declarations, or any combination thereof.

`sem_check_param_decl` is missing this check, I guess because any class was valid in 1993. E.g. `sem_check_port_decl` does

```c
   ...
   else if (class != C_SIGNAL)
      sem_error(t, "invalid object class %s for port %s",
                class_str(class), istr(tree_ident(t)));
```

NikLeberg:
Thanks for the review. Implemented the suggested change and merged the special case for the first port into the loop.

nickg (approved):

Issue #1091, opened 2024-12-05
[Fuzz] Second iteration


avelure:
Following are 13 new crashes from an earlier fuzzing campaign.
For the files with `_standalone` the hash does not match as they had some library dependences and I have minimized them and removed the dependence.

Tested using `nvc 1.15-devel (6ecfc8f) (Using LLVM 15.0.7)` on ubuntu 22.04

[crashes_nvc_new_2.zip](https://github.com/user-attachments/files/18021296/crashes_nvc_new_2.zip)

```
/work/crashes_nvc_new_2/007d677403eec992d943bcd45b40bcac4d961ab3002e10253a47e93374236e59 :
input buffer overflow, can't enlarge buffer because scanner uses REJECT
/work/crashes_nvc_new_2/0c1d01dd6f3aea51f8b086c4c788132486f2bfe6079eb33e89f4e69d39945bc7 :
nvc: ../src/tree.c:998: tree_t tree_ref(tree_t): Assertion `item->object != NULL' failed.

*** Caught signal 6 (SIGABRT) ***

[0x555555613a59] /nvc/build/../src/util.c:900 signal_handler
-->    show_stacktrace();
[0x7ffff0b1351f] (/usr/lib/x86_64-linux-gnu/libc.so.6) 
[0x7ffff0b679fc] (/usr/lib/x86_64-linux-gnu/libc.so.6) pthread_kill
[0x7ffff0b13475] (/usr/lib/x86_64-linux-gnu/libc.so.6) raise
[0x7ffff0af97f2] (/usr/lib/x86_64-linux-gnu/libc.so.6) abort
[0x7ffff0af971a] (/usr/lib/x86_64-linux-gnu/libc.so.6) 
[0x7ffff0b0ae95] (/usr/lib/x86_64-linux-gnu/libc.so.6) __assert_fail
[0x555555727f04] /nvc/build/../src/simp.c:0 simp_tree
[0x55555576c231] /nvc/build/../src/object.c:715 object_rewrite
       for (;;) {
-->       object_t *new = (*ctx->post_fn[object->tag])(object, ctx->context);
          if (new == object || (object = object_rewrite(new, ctx)) == NULL)
[0x55555576bee5] /nvc/build/../src/object.c:779 object_rewrite
                object_t *o = object->items[n].object;
-->             object->items[n].object = object_rewrite(o, ctx);
                object_write_barrier(object, o);
[0x55555576bfdd] /nvc/build/../src/object.c:790 object_rewrite
[...]

PR #1092, opened 2024-12-08
[Fuzz] Raise error when entity aspect is missing in configuration specification.

NikLeberg:
Based on crashing fuzzer input `0c1d01dd6f3aea51f8b086c4c788132486f2bfe6079eb33e89f4e69d39945bc7` from issue #1091.

Previously nvc allowed entity aspect to be non existent even though for such explicit cases I think even the standard requires it to be existent.

> LRM 19 section 7.3.2.1: When a binding indication is used in an explicit configuration specification, it is an error if the entity aspect is absent.

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -9076,7 +9076,11 @@ static void p_configuration_specification(tree_t parent)
 
    push_scope(nametab);
 
+   bool is_open = (peek() == tUSE && peek_nth(2) == tOPEN);
    tree_t bind = p_binding_indication(comp);
+   if (!is_open && bind == NULL)
+      parse_error(CURRENT_LOC, "a binding indication in an explicit "
+                  "configuration specification must contain an entity aspect");
    consume(tSEMI);
 
    if (ids != NULL) {
--- a/test/parse/config.vhd
+++ b/test/parse/config.vhd
@@ -74,3 +74,15 @@ configuration use_pack_ent of ent is
         end for;
     end for;
 end configuration;
+
+entity b is
+end entity b;
+
+architecture rtl of b is 
+  component ent
+  end component;
+
+  for ent0 : ent; -- Error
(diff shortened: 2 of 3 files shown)


nickg:
I would check if `bind != NULL` in `p_configuration_specification` instead. The code would be simpler and the user gets a better error message:

```c
tree_t bind = p_binding_indication(comp);
if (bind == NULL)
   parse_error(CURRENT_LOC, "a binding indication in an explicit configuration "
               "specification must contain an entity aspect");
```

NikLeberg:
The else branch in `p_binding_indication` prevents that `bind` will ever be `NULL` for the case that we care about.
https://github.com/nickg/nvc/blob/4062b258db18b1ddd9add60fb4ff8e44995debea/src/parse.c#L9021-L9022

Otherwise that would be a nice and easy fix.

Up to now `p_binding_indication` returns three kinds of values:
 - `for ent0 : ent;` => _shallow_ tree obj i.e. a simple `tree_new(T_BINDING)`, nothing else
 - `for ent0 : ent use entX;` => a _valid_ tree obj i.e. with location, reference, identifier etc. 
 - `for ent0 : ent use open;` => simply `NULL`

I tried to work around this with the new commit. First I removed the mentioned else branch as it seemed to have no purpose. (All tests pass regardless.) This gives us `NULL` for the first and also the last input case. We then differentiate one from the other by checking if we have a `tOPEN`.

I'm not sure if that is acceptable. I sadly don't know the downstream changes that the removal of the else branch introduces.

NikLeberg:
Nevermind. The removal of the else branch most probably broke something called _incremential binding_. I'll come up with a MWE for that to see if it worked beforehand. If you happen to have a MWE pleaee share.

nickg (approved):

PR #1093, opened 2024-12-08
[Fuzz] Raise error when formal part is illegal.

NikLeberg:
Based on crashing fuzzer input `51c4a1798e1f1aeff3711fc660cc9fa9a06c8ab38777c4fb40e79b23e5682add_standalone` from issue #1091.

The added check ensures that only _legal_ formal parts are parsed / present in the tree. Otherwise attribute references could end up in there.

I don't know what really constitutes as _legal_ formal part. I simply ran the testsuite and added all occurring (and seemingly valid) kinds to the switch case block. All others trigger a new error.
```c
   case T_RECORD_REF:
   case T_ARRAY_REF:
   case T_ARRAY_SLICE:
   case T_TYPE_CONV:
      break;
```
Please let me know if I forgot something.

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -3123,8 +3123,15 @@ static tree_t p_formal_part(type_t *signature)
       }
       break;
 
-   default:
+   case T_RECORD_REF:
+   case T_ARRAY_REF:
+   case T_ARRAY_SLICE:
+   case T_TYPE_CONV:
       break;
+
+   default:
+      parse_error(CURRENT_LOC, "illegal formal designator");
+      return error_expr();
    }
 
    return name;
--- a/src/parse.c
+++ b/src/parse.c
@@ -3109,6 +3109,12 @@ static tree_t p_formal_part(type_t *signature)
    tree_t name = p_name(0);
 
    switch (tree_kind(name)) {
+   case T_RECORD_REF:
+   case T_ARRAY_REF:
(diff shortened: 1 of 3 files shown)


NikLeberg:
Changed to WIP because crash `4458d32c108c23646b56560b972a18c380feb23e29fd65aab776d729952e77c6_standalone` is hitting the same error but from another codepath and this fix is not fixing it.

NikLeberg:
Added a check for when the formal part is of kind `T_REF` but the name could not be resolved i.e. the type is `T_NONE`. That was the only (easy) difference I could make out between valid and invalid `T_REF` formal part tree entries. In such cases the same newly added error `invalid formal designator` is triggered.

src/parse.c
```diff
@@ -3121,10 +3127,17 @@ static tree_t p_formal_part(type_t *signature)
...
+      // Fall-through
 
    default:
-      break;
+      parse_error(CURRENT_LOC, "illegal formal designator");
+      tree_t dummy = tree_new(T_REF);
+      tree_set_loc(dummy, CURRENT_LOC);
+      tree_set_ident(dummy, error_marker());
+      tree_set_type(dummy, type_new(T_NONE));
+      return dummy;
```
nickg, line 3140:
```suggestion
      return error_expr();
```

nickg (approved):

PR #1099, opened 2024-12-12
[Fuzz] Show fuzzy-stack-hash for crashes in debug build.

NikLeberg:
Hi there,

this is something that helped me to quickly see if different reproducers result in the same crash. Source of the idea is: https://argp.github.io/2014/12/29/fuzzy-stack-hash/

Basically the symbol names of the N last stack frames of the backtrace are hashed. Usually for crashes the text _Please report this bug_ is shown. For debug builds this text is omitted. This PR would add the output of a crash hash in the following form for debug builds:
```
<other stack frames>
[0x565027925c4a] ../src/nvc.c:2193 process_command
       case 'a':
-->       return analyse(argc, argv, state);
       case 'e':
[0x56502792618d] ../src/nvc.c:2355 main
-->    const int ret = process_command(argc, argv, &state);

Crash hash: 79505df7
```

Please feel no obligation to merge this. I simply wanted to offer it. If you deem it useful, you are of course more than welcome to merge it.

Cheers

### DIFF
--- a/src/debug.h
+++ b/src/debug.h
@@ -61,6 +61,7 @@ debug_info_t *debug_capture(void);
 void debug_free(debug_info_t *di);
 unsigned debug_count_frames(debug_info_t *di);
 const debug_frame_t *debug_get_frame(debug_info_t *di, unsigned n);
+uint32_t debug_hash(debug_info_t *di);
 
 void debug_add_unwinder(void *start, size_t len, debug_unwind_fn_t fn,
                         void *context);
--- a/src/util.c
+++ b/src/util.c
@@ -713,6 +713,10 @@ void show_stacktrace(void)
 
    }
 
+#ifdef DEBUG
+   color_fprintf(stderr, "\n$!red$Crash hash: %08x$$\n", debug_hash(di));
+#endif
+
    debug_free(di);
 
 #if defined __linux__ && !defined HAVE_LIBDW && !defined HAVE_LIBDWARF
--- a/src/debug.c
+++ b/src/debug.c
@@ -62,6 +62,7 @@
 #endif
 
 #define MAX_TRACE_DEPTH   25
+#define MAX_HASH_DEPTH    10
 
 struct debug_info {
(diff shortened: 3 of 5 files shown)


src/util.c
```diff
@@ -713,6 +713,10 @@ void show_stacktrace(void)
 
    }
 
+#ifdef DEBUG
+   color_fprintf(stderr, "\n$!red$Crash hash: %08x$$\n", debug_hash(di));
+#endif
```
nickg, line 718: Can this be `#ifdef FUZZER_IS_ENABLED` or something like that?

src/debug.h
```diff
@@ -61,6 +61,7 @@ debug_info_t *debug_capture(void);
 void debug_free(debug_info_t *di);
 unsigned debug_count_frames(debug_info_t *di);
 const debug_frame_t *debug_get_frame(debug_info_t *di, unsigned n);
+hash_state_t debug_hash(debug_info_t *di);
```
nickg, line 64: I think this can just return an opaque `uint32_t`

nickg (commented):
I'm ok with this but I'm a bit nervous that moving the hashing code to `util.c` means it won't be inlined into `ident_new`, etc. any more (I haven't checked whether LTO does this or not). Since it's only a few lines of code I'd rather just copy and paste it into `debug_hash` directly.

NikLeberg:
> Can this be `#ifdef FUZZER_IS_ENABLED` or something like that?

I don't want to pollute your codebase with fuzzing-only code. If we are gonna hide the crash hash behind a fuzzing specific flag, then I can also just put this feature as a manual patch file in [NikLeberg/nvc-fuzz](https://github.com/NikLeberg/nvc-fuzz).

I'll close this for now.

BTW, if you put `-Winline`, then GCC will warn about functions marked with `inline` that it did not inline. If I do that, then it warns, that e.g. `p_expression`, `lvn_can_fold` or `split_nexus` cannot be inlined. There are also others that were not inlined.

PR #1104, opened 2024-12-14
[Fuzz] Fix crash with missing choices in aggregate.

NikLeberg:
Based on crashing fuzzer input `da346642705230c6f2c89b7c0a522d1078c44b3d0dfd5fe52e2859212edd2784` from issue #1091.

The crash happens because `simplify_local` cannot find the missing choice in the aggregate. Bounds checking reports an error for it, but the check is only done afterwards. This adds a check for missing aggregate choices in `eval_possible` and forbids the early evaluation of such erroneous constructs.

Alternatively one could also do the bounds check before the local simplify and only if there were no errors, but I think that would change the code flow in a lot of places and tests.

Cheers

### DIFF
--- a/src/eval.c
+++ b/src/eval.c
@@ -364,6 +364,65 @@ bool eval_possible(tree_t t, unit_registry_t *ur)
                return false;
          }
 
+         // Check for missing choices in constrained array aggregates
+         type_t composite_type = tree_type(t);
+         if (type_is_array(composite_type)
+             && !type_is_unconstrained(composite_type)) {
+            int64_t count = 0, elem_count = 0;
+            bool known_elem_count = false;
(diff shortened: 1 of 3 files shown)


nickg:
> Alternatively one could also do the bounds check before the local simplify and only if there were no errors, but I think that would change the code flow in a lot of places and tests.

The problem with that is some bounds errors can only be detected after you've folded locally static expressions. I agree the current order is not great either as it means invalid VHDL ends up in the code generator (this isn't the first case like this either). 

Adding a defensive check here is fine. Perhaps a better organisation would be to merge the simplify and bounds checks passes rather than doing them one after another. I'll take a look at this.

nickg (approved):

PR #1106, opened 2024-12-17
[Fuzz] Fix crash with duplicate aliases in name resolution.

NikLeberg:
When we have more than one alias of the same name but both with no valid type, then we crash in `get_result_type` that gets called from `try_resolve_name`.

I'm not sure what the implications for this change are. All tests pass though.
Alternatively I did come up with another (less elegant) fix that would at least set the type to `T_NONE` for both the invalid aliases:
```diff
diff --git a/src/parse.c b/src/parse.c
index 8a0ed211..ce115359 100644
--- a/src/parse.c
+++ b/src/parse.c
@@ -7471,18 +7471,17 @@ static void p_alias_declaration(tree_t parent)
       }
       tree_set_type(t, type);
    }
-   else if (value_kind == T_REF && tree_has_ref(value)) {
+   else if (value_kind == T_REF && tree_has_ref(value)
+            && is_design_unit(tree_ref(value))) {
       // A nonobject alias may be a design unit which does not have a type
-      tree_t decl = tree_ref(value);
-      if (is_design_unit(decl)) {
-         tree_set_flag(t, TREE_F_NONOBJECT_ALIAS);
-         nonobject_alias = true;
-      }
-      else
-         solve_types(nametab, value, NULL);
+      tree_set_flag(t, TREE_F_NONOBJECT_ALIAS);
+      nonobject_alias = true;
+   }
+   else {
+      type_t type = solve_types(nametab, value, NULL);
+      if (type_is_none(type))
+         tree_set_type(t, type);
    }
-   else
-      solve_types(nametab, value, NULL);
 
    if (value_kind == T_REF && tree_has_ref(value)) {
       tree_t decl = tree_ref(value);
```

Cheers!

### DIFF
--- a/src/names.c
+++ b/src/names.c
@@ -1611,6 +1611,9 @@ static void hint_for_typo(scope_t *top_scope, diag_t *d, ident_t name,
 
 static type_t get_result_type(nametab_t *tab, tree_t decl)
 {
+   if (tree_kind(decl) == T_ALIAS && !tree_has_type(decl))
+      return tree_type(tree_value(decl));
+
    type_t type = tree_type(decl);
    if (type_kind(type) != T_SIGNATURE || !type_has_result(type))
       return type;
--- a/test/parse/alias5.vhd
+++ b/test/parse/alias5.vhd
@@ -1,0 +1,5 @@
+package test_pkg is
+  alias alias_t is type_t;
+  alias alias_t is type_t;
+  constant c : alias_t;
+end package;
--- a/test/test_parse.c
+++ b/test/test_parse.c
@@ -7089,6 +7089,24 @@ START_TEST(test_issue1096)
 }
 END_TEST
 
+START_TEST(test_alias5)
+{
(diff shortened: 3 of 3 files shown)


src/names.c
```diff
@@ -1611,6 +1611,9 @@ static void hint_for_typo(scope_t *top_scope, diag_t *d, ident_t name,
 
 static type_t get_result_type(nametab_t *tab, tree_t decl)
 {
+   if (!tree_has_type(decl))
+      return type_new(T_NONE);
+
```
nickg, line 1616: I think this works too

```suggestion
   if (tree_kind(decl) == T_ALIAS && !tree_has_type(decl))
      return tree_type(tree_value(decl));

```

It makes a bit more sense when reading it ("if the thing is an alias declaration and it didn't have a type indication then use the type of aliased object (which will be `T_NONE` here").

NikLeberg: Agreed, that conveys more meaning and also fixes the crash.

nickg (approved):

PR #1110, opened 2024-12-19
make --define work also for verilog preprocessor macros.

Blebowski:
Insert macros from `--define` option also to Verilog processor.

Do not de-initialize the macro table after single Verilog file is pre-processed.
This matches Verilog style behavior where all files compiled via single invocation
of tool are considered as single compilation unit.


nickg:
I did this slightly differently in 2ce462d.

> This matches Verilog style behavior where all files compiled via single invocation of tool are considered as single compilation unit.

The System Verilog LRM actually allows either behaviour, see 3.12.1:

> The exact mechanism for defining which files constitute a compilation unit is tool-specific. However, compliant tools shall provide use models that allow both of the following cases:
> a) All files on a given compilation command line make a single compilation unit (in which case the declarations within those files are accessible following normal visibility rules throughout the entire set of files).
> b) Each file is a separate compilation unit (in which case the declarations in each compilation-unit scope are accessible only within its corresponding file).

I don't particularly like the the mode where all files are a single compilation unit and I'd rather not have it by default, but I added a `--single-unit` option to enable this.

PR #1126, opened 2025-01-08
Add support for `'*_name` attributes of generate block label

### LINKED ISSUE #1125: Error: no visible declaration for LABEL
The following code

```VHDL
  gen_logic : IF (g_test = 1) GENERATE
  
  ...

    ASSERT NOT (i_rdreq = '1')
      REPORT "Read - " & GEN_LOGIC'PATH_NAME & " @ " & time'IMAGE(now)
      SEVERITY warning;

  ...

  END GENERATE gen_logic;

```

generates an error

```
** Error: no visible declaration for GEN_LOGIC
     > file.vhd:181
     |
 181 |       REPORT "Read - " & gen_logic'PATH_NAME & " @ " & time'IMAGE(now)
```

The code works for modelsim without an error.

NikLeberg:
I had a bit of fun trying to fix issue #1125 as it seemed rather simple.

The basic stuff works. But some other does not:
 - [ ] does not support alternative labels for if/case generate blocks
   - I had trouble supporting both, the normal label and the alternative label at the same time. Because the current code overwrites the ident of the generate block if an alternative label was found. In such a case `gen_case` and `gen_case_true` would both resolve to `gen_case_true`.
 - [ ] if the `'path_name` is accessed from outside the generate block, instead of returning e.g. `:issue1125:gen_case:` it returns only `:issue1125:`
   - Interesting to think about what we would even return for the `gen_for` label if we are outside the loop. Inside we have `gen_for(0)`, `gen_for(1)`, etc. But outside? Only `gen_for`?
   - I tested this with modelsim and it returns whatever it wants.... sometimes `gen_for(1)` and sometimes `gen_for`. So it does not seem very important.

Let me know if you see another way of implementing it. Please feel also absolutely free to discard everything and implement your own fix.

Cheers

Fixes #1125

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -42,6 +42,8 @@
 - `vhpi_assert` now behaves the same as VHDL `assert` for the purposes
   of determining the simulation exit code and early termination (#1060).
 - The `'driving_value` attribute now works correctly with record types.
+- Added basic support for `'instance_name`, `'path_name` and `'simple_name`
+  attributes of generate block labels (from @NikLeberg) (#1125).
 - Several other minor bugs were resolved (#1038, #1057, #1067).
 
 ## Version 1.14.2 - 2024-11-23
--- a/src/common.c
+++ b/src/common.c
@@ -574,6 +574,9 @@ class_t class_of(tree_t t)
    case T_CONCURRENT:
    case T_ELAB:
    case T_PSL:
+   case T_FOR_GENERATE:
+   case T_IF_GENERATE:
+   case T_CASE_GENERATE:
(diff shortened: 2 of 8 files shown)


nickg (approved):

nickg:
> does not support alternative labels for if/case generate blocks 

I need to check what the LRM says here but I think this is fine as-is. 

> if the 'path_name is accessed from outside the generate block, instead of returning e.g. :issue1125:gen_case: it returns only :issue1125:

Concurrent blocks have the same problem. I'll try to fix it at the weekend.

PR #1132, opened 2025-01-13
Replace redundant copy of libraryName with missing copy of modelName

svnesbo:
I stumbled over this earlier, figured it must be a mistake?


nickg:
> I stumbled over this earlier, figured it must be a mistake?

Yes looks like it, thanks.

PR #1139, opened 2025-01-19
Extend PSL prev with configurable cycle count.

Blebowski:
Extends PSL `prev()` with second argument that is higher than 1.

The implementation is naive. A "shift register" is maintained for each `prev()` call.
Depth of shift register is given by value of `Number`. Upon even update, the
shift register is shifted towards LSB. LSB is the result value of the function call.

A better implementation would have ring buffer with head pointer. I assume that
the `Number` value is small (couple of cycles at most) so amount of copying is not
excessive.

### DIFF
--- a/src/psl/psl-lower.c
+++ b/src/psl/psl-lower.c
@@ -289,19 +289,19 @@ vcode_reg_t psl_lower_fcall(lower_unit_t *lu, psl_node_t p)
    tree_t expr = psl_tree(psl_operand(p, 0));
    type_t type = tree_type(expr);
 
-   if (psl_operands(p) > 1) {
-      const int num = assume_int(psl_tree(psl_operand(p, 1)));
-      if (num != 1)
-         fatal_at(psl_loc(p), "sorry, cycle counts other than 1 are "
-                  "not supported");
-   }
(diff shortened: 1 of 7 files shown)


src/psl/psl-lower.c
```diff
@@ -289,19 +289,19 @@ vcode_reg_t psl_lower_fcall(lower_unit_t *lu, psl_node_t p)
...
-   }
+   int num = 1;
+   if (psl_operands(p) > 1)
+      num = assume_int(psl_tree(psl_operand(p, 1)));
 
    vcode_type_t vtype = lower_type(type);
    vcode_type_t vbounds = lower_bounds(type);
 
-   vcode_var_t var = emit_var(vtype, vbounds, ident_uniq("prev"), 0);
+   vcode_var_t vars[num];
```
nickg, line 299:
```suggestion
   vcode_var_t *vars LOCAL = xmalloc_array(num, sizeof(vcode_var_t));
```

It's better to allocate on the heap when the size is user controlled. I think we should also enforce a sensible upper limit here like 512.

nickg (approved):

PR #1149, opened 2025-01-27
src: add run-time randomizer and implement PSL union with it.

Blebowski:
Implement simple run-time randomizer as 64 bit LFSR. Held with RT model
and seeded by additional CLI argument `--seed`. If not given a random seed
is taken from system time.

Lowering of PSL union implemented to pick either of its arguments based
on random value queried from RT model via `emit_get_random`.

The `union` parsing was reworked to be allowed at arbitrary place in VHDL expression (similar to built-in function call).

A simple test was added checking random selection of "bit", "bit_vector" and "integer" arguments.
A negative test for error when union with two distinct arguments is used was added.


Blebowski:
any opinion on this @nickg ?

Blebowski:
Hi @nickg ,

do you have any preference about these changes ? Should I close the MR, or keep rebasing it every now and then ?

nickg:
Thanks. I split this up a bit and used a different random number generation scheme.

PR #1160, opened 2025-02-11
Parse ANSI parameters declaration

Anselmo95:
Add changes to parse parameters inside `#()` for a module declaration.

### DIFF
--- a/src/lexer.l
+++ b/src/lexer.l
@@ -139,6 +139,7 @@ UDP_INDICATOR   "("{UDP_LEVEL}{UDP_LEVEL}")"
 SCALAR_ZERO     ('b0|'B0|1'b0|1'B0)
 SCALAR_ONE      ('b1|'B1|1'b1|1'B1)
 PAREN_STAR      "("{SPACE}*"*"{SPACE}*")"
+HASH_LPAREN     #{SPACE}*"("
 
 %x COMMENT C_COMMENT PSL VLOG UDP SDF SDF_EXPR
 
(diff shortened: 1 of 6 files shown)


src/lexer.l
```diff
@@ -505,6 +505,7 @@ BEFORE            ?i:before
 <UDP>{UDP_EDGE}          { yylval.i64 = yytext[0]; return tUDPEDGE; }
 <UDP>{UDP_INDICATOR}     { yylval.str = xstrdup(yytext); return tUDPIND; }
 
+<*>"#("                  { TOKEN(tHASHLPAREN); }
```
nickg, line 508: I don't think `#(` is a single token. It's shown with a space between it them the grammar in the standard (e.g. page 1138 of 1800-2017) and also here:

https://www.sigasi.com/tech/systemverilog.ebnf/#module-parameters-and-ports

Icarus Verilog and Verilator also both treat the `#` and `(` as separate tokens.

src/vlog/vlog-parse.c
```diff
@@ -3260,6 +3260,33 @@ static void p_list_of_port_declarations(vlog_node_t mod)
...
+
+   consume(tPARAMETER);
+
+   vlog_node_t dt = p_data_type_or_implicit();
+   vlog_add_decl(mod, p_param_assignment(dt));
+}
+
+static void p_list_of_param_declarations(vlog_node_t mod)
+{
+   // parameter_port_list ::= ansi_param_assignments { , ansi_param_assignments }
```
nickg, line 3277: I'm using the System Verilog grammar rather than the traditional Verilog one

src/lexer.l
Anselmo95: Ok thanks for the review I wasn't aware that they were separate tokens, I'll try to look better into the syntax and adjust the code. It's gonna take a bit of time though

nickg (approved):

PR #1166, opened 2025-02-28
Update TCL channels and fix ARM_CRYPTO attributes

kammoh:
These changes are required for compiling under macOS/M1.
I'm really addressing two different issues, so let me know if you'd like this to be split into two PRs.

### DIFF
--- a/src/rt/shell.c
+++ b/src/rt/shell.c
@@ -1111,7 +1111,8 @@ static char *shell_list_generator(const char *script, const char *text,
                                   int state, int prefix)
 {
    static Tcl_Obj *list = NULL;
-   static int index, len, max;
+   static int index, len;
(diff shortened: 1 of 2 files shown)


src/rt/shell.c
```diff
@@ -1111,7 +1111,8 @@ static char *shell_list_generator(const char *script, const char *text,
                                   int state, int prefix)
 {
    static Tcl_Obj *list = NULL;
-   static int index, len, max;
+   static int index, len;
+   static Tcl_Size max;
```
nickg, line 1115:
```suggestion
#if TCL_MAJOR_VERSION >= 9
   static Tcl_Size max;
#else
   static int max;
#endif
```

There's no `Tcl_Size` in TCL 8.

PR #1206, opened 2025-05-24
Cover embed source files

Blebowski:
This MR copies each source file to a stand-alone HTML file when generating code coverage report.
The generated HTML file is then referenced from the "File" in the coverage hierarchy header.
This is usefull when e.g. a regression is run in CI, and code coverage report is extracted in artifacts.
After downloading the artifacts, the reference to the source files remains valid.

To do this, I needed to track location of the underlying "unit" for the scope, not the "instance".
It affects the XML file output, but I guess this is no big deal since this format is subject to change.
The issue909 also now gives more relevant output since the source file for instance itself can't be found.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,8 @@
   previous behaviour.
 - Fixed a crash when `set_cover_scope_name` from `nvc.cover_pkg` was
   called with coverage disabled (#1218).
+- Code coverage reports now contain the source files instead of
+  referencing them.
 
 ## Version 1.16.2 - 2025-06-11
 - Fixed a crash when an entity uses VHDL-2019 enhanced type generics and
--- a/src/cov/cov-data.c
+++ b/src/cov/cov-data.c
@@ -999,7 +999,7 @@ cover_scope_t *cover_create_instance(cover_data_t *data, cover_scope_t *parent,
    cover_scope_t *s = xcalloc(sizeof(cover_scope_t));
    s->name       = tree_ident(block);
    s->parent     = parent;
-   s->loc        = *tree_loc(block);
+   s->loc        = *tree_loc(unit);
    s->hier       = ident_prefix(s->parent->hier, s->name, '.');
    s->emit       = cover_should_emit_scope(data, s, block);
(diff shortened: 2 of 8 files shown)


sean-anderson-seco:
This places source files in the current working directory instead of under `--output`. The links are broken because of this too.

Something like

```
cd build
nvc --work=test:tb/test -Lrtl -a ../tb/foo.vhd
nvc --work=test:tb/test -Lrtl -e --cover --cover-file=tb/test.sim/foo.ncdb foo
nvc --work=test:tb/test -Lrtl -r foo
nvc --cover-report --output=tb/test.cov tb/test.sim/foo.ncdb
```

produces something like

```
build/3ae69fb2889865e561bc946caa89643ad9daa5b6.html
build/tb/test.cov/
build/tb/test.cov/hier
build/tb/test.cov/hier/5bb7fd2eb53977f94e1bdcaf1d2569c7f09d6344.html
build/tb/test.cov/index.html
```

and the link in `5bb7.html` is to `../../3ae6.html` which is invalid since it points at `build/tb/3ae6.html`.

The files produced should instead be something like

```
build/
build/tb/test.cov/
build/tb/test.cov/hier
build/tb/test.cov/hier/5bb7fd2eb53977f94e1bdcaf1d2569c7f09d6344.html
build/tb/test.cov/src/3ae69fb2889865e561bc946caa89643ad9daa5b6.html
build/tb/test.cov/index.html
```

nickg:
> This places source files in the current working directory instead of under --output. The links are broken because of this too.

AFAIK this was fixed in 370560a.

sean-anderson-seco:
That fixes files getting placed in the working directory, but the links are still broken. They point at `../../tb/test.cov/src/3ae6.html` which is `build/tb/tb/test.cov/src/3ae6.html`.

nickg:
> That fixes files getting placed in the working directory, but the links are still broken. They point at `../../tb/test.cov/src/3ae6.html` which is `build/tb/tb/test.cov/src/3ae6.html`.

OK, I think this is fixed too now.

PR #1217, opened 2025-06-16
Add `--ieee-warnings=off-at-0` option

### LINKED ISSUE #1214: Feature request: `off-at-0` option for `--ieee-warnings`
First off, awesome simulator, the error stack traces are wonderful!

I found NVC after trying GHDL, and in my testbenches, I usually leave designs uninitialised and later use a reset to check it works. As expected, this results in many errors from the IEEE packages about metastable values.

The GHDL option [`--ieee-asserts`](https://ghdl.github.io/ghdl/using/Simulation.html#cmdoption-ghdl-ieee-asserts) has an option `disable-at-0` which I found really helpful instead of disabling all IEEE warnings. The respective NVC option [`--ieee-warnings`](https://www.nickg.me.uk/nvc/manual.html#ieee-warnings=) only has `on` or `off`.

I tried implementing the feature myself and I think I got it working, should I open a PR adding `off-at-0` as a value for this option?

Edit: Made the PR since I don't think it would need much back-and-forth

supleed2:
Closes #1214

## Changes:

- Add an enum for the `ieee-warnings` options (updating `0` / `1` constants with enum values), and a parse function (based on the previous `parse_on_off`)
- Add a check to `_nvc_ieee_warnings` that returns `IEEE_WARNINGS_OFF` if it is set to `off-at-0` and time is 0
- Update help messages
- Update standard library modules: `NO_WARNING` constant to a function so it is evaluated when needed
- I believe my changes follow [contrib/STYLE.md](contrib/STYLE.md), but point out if I've missed something :) 

I also noticed when compiling to test: `parse_on_off` was only used for this option, so it is now unused, should I remove it?

## Testcase:

Command: `nvc --ieee-warnings=off-at-0 -a tb.vhdl --check-synthesis -e tb -r --format fst --wave`

`tb.vhdl`:

```vhdl
library ieee;
use ieee.std_logic_1164.all;
use ieee.numeric_std.all;

entity tb is
end;

architecture sim of tb is
begin
    process
        variable v : STD_LOGIC_VECTOR (7 downto 0) := x"XX";
        variable i : INTEGER;
    begin
        report "start" severity NOTE;
        i := TO_INTEGER(UNSIGNED(v));
        wait for 1 ns;
        report "wait 1 ns" severity NOTE;
        i := TO_INTEGER(UNSIGNED(v));
        report "end" severity NOTE;
        wait;
    end process;
end;
```

### DIFF
--- a/lib/ieee.08/numeric_bit-body.vhdl
+++ b/lib/ieee.08/numeric_bit-body.vhdl
@@ -66,7 +66,10 @@ package body NUMERIC_BIT is
 
   -- implementation controls
 
-  constant NO_WARNING : BOOLEAN := not ieee_warnings;
+  function NO_WARNING return BOOLEAN is
+  begin
+    return not ieee_warnings;
+  end NO_WARNING;
 
   -- =========================Local Subprograms =================================
 
--- a/lib/ieee.08/numeric_std-body.vhdl
+++ b/lib/ieee.08/numeric_std-body.vhdl
@@ -74,7 +74,10 @@ package body NUMERIC_STD is
 
   -- implementation controls
 
-  constant NO_WARNING : BOOLEAN := not ieee_warnings;
+  function NO_WARNING return BOOLEAN is
+  begin
+    return not ieee_warnings;
+  end NO_WARNING;
 
   -- =========================Local Subprograms =================================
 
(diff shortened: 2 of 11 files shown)


nickg:
I pushed a modified version of this in 8095f0c. The problem with making `ieee_warnings` a function call is that it significantly increases the overhead in the common case of `--ieee-warnings=off` which used to be just a load from memory before.  Instead it seems better to just modify the value of the option/constant after the first time step.

supleed2:
Oh awesome, I had considered the overhead of changing the VHDL constant to an function call, but didn't know how I would go about modifying a constant from within the simulator, thanks!

PR #1228, opened 2025-07-01
Fixup  initial/fork blocks and empty function/task calls

Anselmo95:
- Fix parsing  of labels and error check them for initial/fork blocks
- Fix function/tasks definitions and calls when parenthesis are used with no arguments

### DIFF
--- a/src/vlog/vlog-parse.c
+++ b/src/vlog/vlog-parse.c
@@ -1266,15 +1266,16 @@ static void p_list_of_arguments(vlog_node_t call)
 
    BEGIN("list of arguments");
 
-   do {
-      if (peek() == tCOMMA) {
-         vlog_node_t v = vlog_new(V_EMPTY);
-         vlog_set_loc(v, &state.last_loc);
-         vlog_add_param(call, v);
-      }
-      else
-         vlog_add_param(call, p_expression());
-   } while (optional(tCOMMA));
+   if (peek() != tRPAREN)
+      do {
+         if (peek() == tCOMMA) {
+            vlog_node_t v = vlog_new(V_EMPTY);
+            vlog_set_loc(v, &state.last_loc);
+            vlog_add_param(call, v);
+         }
+         else
+            vlog_add_param(call, p_expression());
+      } while (optional(tCOMMA));
 }
 
 static vlog_node_t p_system_tf_call(vlog_kind_t kind)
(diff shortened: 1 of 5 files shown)


src/vlog/vlog-parse.c
```diff
@@ -1894,10 +1894,20 @@ static vlog_node_t p_seq_block(void)
...
    if (optional(tCOLON))
       vlog_set_ident(v, p_identifier());
 
+   // TODO: need a way to peek past attrubutes to know if a
+   // block_item_declaration or a statement_or_null follows
+   // the following code cannot work properly with attributes
+   while (scan(tREG, tSTRUCT, tUNION, tTYPEDEF, tENUM, tSVINT, tINTEGER,
+               tSVREAL, tSHORTREAL, tREALTIME, tBIT, tLOGIC))
+      p_block_item_declaration(v);
+
```
Anselmo95, line 1910: I wanted to keep the attributes parsing inside **p_block_item_declaration** but I  could not find a way to do the scan there without considering attributes. For these cases it would be nice to have a version of peek that returns the token after all attributes

src/vlog/vlog-parse.c
```diff
@@ -5498,6 +5518,10 @@ static vlog_node_t p_description(void)
    BEGIN("description");
 
    switch (peek()) {
+   case tATTRBEGIN:
+      while (peek() == tATTRBEGIN)
+         p_attribute_instance();
+      return p_description();
```
Anselmo95, line 5524: I added this to parse attributes before a module but as in the other comment it would be nice to peek after the attributes for this switch

src/vlog/vlog-parse.c
nickg: Since we just ignore attributes we can skip over them before calling `scan()`. See f2ad37996ef428a08.

test/vlog/initial1.v
```diff
@@ -0,0 +1,17 @@
+(* attr0 *) 
+(* attr1 *) 
+(* attr2 *) 
+module initial1;
+  initial begin : label0
+    reg w;
+    w = 1;
+    $display("ok");
+    reg q = 0; // Is this allowed ?
```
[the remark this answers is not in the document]
Anselmo95: OK thanks for the confirmation. I seem an error according to the specs but I was not so sure

src/vlog/vlog-parse.c
```diff
@@ -1306,7 +1306,8 @@ static vlog_node_t p_tf_call(vlog_kind_t kind)
    vlog_set_ident(v, p_identifier());
 
    if (optional(tLPAREN)) {
-      p_list_of_arguments(v);
+      if (peek() != tRPAREN)
```
nickg, line 1309: I don't think this is correct. The grammar says `[ ( list_of_arguments ) ]` i.e. `list_of_arguments` is not optional if there is a `(`. This check should be inside `p_list_of_arguments` (because the first `expression` there is optional).

Anselmo95: Ah you are right, thanks for the correction I'll fix it asap

nickg (approved):

PR #1244, opened 2025-07-16
Parse unconnected_drive and nounconnected_drive directives

Anselmo95:


src/vlog/vlog-parse.c
```diff
@@ -6151,11 +6180,11 @@ vlog_node_t vlog_parse(void)
       vlog_symtab_poison(symtab, error_marker());
    }
 
+   p_directive_list();
+
    if (peek() == tEOF)
       return end_of_file();
 
-   p_directive_list();
-
```
Anselmo95, line 6158: I moved the check for the end of file after the p_directive_list otherwise there is a crash when a directive is the last thing on a file. But I am not too sure if this is correct, it was a quick fix i found just to avoid the crash

nickg: I think this is OK.

nickg (approved):

src/vlog/vlog-parse.c
Anselmo95: Ok, thanks for the answer and merging all my PR !

PR #1258, opened 2025-08-03
Fix 'converse for nested views

rcgoodfellow:
IEEE 1076-2019 section 16.2.8 says that applying `'CONVERSE` to a mode view means that each element mode `EM` becomes `EM'CONVERSE`.  This PR attempts to bring NVC in line with that part of the VHDL 2019 standard.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -514,8 +514,31 @@ tree_t find_element_mode_indication(tree_t view, tree_t field, bool *converse)
          const int nelems = type_fields(view_type);
          for (int i = 0; i < nelems; i++) {
             tree_t e = type_field(view_type, i);
-            if (tree_ref(e) == field)
+            if (tree_ref(e) == field) {
+               // If this element has a view and we need to apply converse,
+               // recursively apply converse to the element's view
+               if (*converse && tree_has_value(e)) {
+                  tree_t elem_view = tree_value(e);
+                  
+                  // Create a 'converse attribute reference for the element's view
+                  tree_t converse_attr = tree_new(T_ATTR_REF);
+                  tree_set_name(converse_attr, elem_view);
+                  tree_set_subkind(converse_attr, ATTR_CONVERSE);
+                  tree_set_loc(converse_attr, tree_loc(elem_view));
+                  
+                  // Create a new element with the converse view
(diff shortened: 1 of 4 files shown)


test/regress/view4.vhd
```diff
@@ -78,14 +78,14 @@ begin
         assert o = (0, 0, ("000", "000"));
         wait for 0 ns;
         assert i = (0, 5, ("000", "000"));
-        assert o = (1, 0, ("111", "111"));
+        assert o = (1, 0, ("000", "111"));
```
nickg, line 81: I don't think this is correct: `o.p.y` should be `out` inside instance `u` which it gets from `vec_in_view'converse` and that in turn from `alias out_view is in_view'converse`.

src/common.c
```diff
@@ -514,8 +514,31 @@ tree_t find_element_mode_indication(tree_t view, tree_t field, bool *converse)
          const int nelems = type_fields(view_type);
          for (int i = 0; i < nelems; i++) {
             tree_t e = type_field(view_type, i);
-            if (tree_ref(e) == field)
+            if (tree_ref(e) == field) {
+               // If this element has a view and we need to apply converse,
+               // recursively apply converse to the element's view
+               if (*converse && tree_has_value(e)) {
```
nickg, line 520: The idea is the caller checks `*converse` or passes it to `converse_mode()`. Of course it's possible that some callers handle it incorrectly.

test/regress/view7.vhd
```diff
@@ -0,0 +1,47 @@
...
+entity view7 is
+    port (
+        p : view converse_outer_view
+    );
+end entity;
+
+architecture test of view7 is
+begin
+    -- This assignment only compiles if converse was applied to nested view
+    p.inner.a <= 42;
```
nickg, line 41: I tried this on the latest version and there's no error (but line 44 *should* be an error but isn't). Was this reduced from a larger example that had incorrect behaviour?

nickg: Thanks, I pushed a fix for this in 293f7f3.

rcgoodfellow: Thanks!

PR #1297, opened 2025-09-20
Parse System Verilog packages

Blebowski:


src/vlog/vlog-parse.c
```diff
@@ -3781,7 +3781,7 @@ static void p_block_item_declaration(vlog_node_t parent)
    }
 }
 
-static void p_package_or_generate_item_declaration(vlog_node_t mod)
+static void p_package_or_generate_item_declaration(vlog_node_t v)
```
nickg, line 3784:
```suggestion
static void p_package_or_generate_item_declaration(vlog_node_t parent)
```

src/vlog/vlog-parse.c
```diff
@@ -6474,6 +6474,44 @@ static vlog_node_t p_udp_declaration(void)
...
+               tLONGINT, tBYTE, tTASK, tFUNCTION, tLOCALPARAM, tPARAMETER))
+      p_package_or_generate_item_declaration(v);
+
+   consume(tENDPACKAGE);
+
+   if (optional(tCOLON)) {
+      ident_t end_name = p_identifier();
+      if (name != end_name)
+         error_at(&state.last_loc, "Trailing identifier '%s' "
+                  "must match the package name '%s'",
```
nickg, line 6508:
```suggestion
         error_at(&state.last_loc, "trailing identifier '%s' "
                  "must match the package name '%s'",
```

Messages always start with a lower case letter.

Or else

```suggestion
         error_at(&state.last_loc, "'%s' does not match package name '%s'",
```

would match the structure of other similar messages.

nickg (approved):

PR #1319, opened 2025-10-06
Partial support of PSL next_a

Blebowski:


Blebowski:
This is now resurected from the dead.

There is one mismatch, in Aldec the message is printed 4 times, while NVC prints it 5 times.
I think Aldec does not print message in the same cycle twice, while NVC does.

Other than that, this is ready for review, upstreaming.

nickg:
> There is one mismatch, in Aldec the message is printed 4 times, while NVC prints it 5 times. I think Aldec does not print message in the same cycle twice, while NVC does.

I think it's caused by the guard check being one cycle too early. I fixed it.

nickg (approved):

Issue #1004, opened 2024-09-30
Feature request: Warning suppression and escalation to error command-line flags


soronpo:
I'm missing common command-line option that:
* Escalate warnings into errors (so that the program returns non-zero error code)
* Completely suppress specific warnings

nickg:
Can you give some examples of the warnings you want suppress and/or those you want to escalate?

soronpo:
Warning suppression example of shared variables that model TrueDPR:
```
** Warning: shared variable RAM must have protected type
    > TrueDPR\hdl\TrueDPR.vhd:28
    |
 28 |   shared variable ram : t_vecX1_std_logic_vector(0 to 2 ** ADDR_WIDTH - 1)(DATA_WIDTH - 1 downto 0);
    |                   ^^^
```
Escalation is a companion to suppression. I strive for 0 warnings build/simulations. So all warnings are either considered errors that must be fixed or alternatively suppressed and completely ignored.

Blebowski:
Related to: https://github.com/nickg/nvc/discussions/784

PR #1343, opened 2025-11-09
add -Werror to escalate warnings to errors.

Blebowski:


Blebowski:
Hi @nickg ,

do you please have any comments to this MR ?

I was hoping to get this in before giving a shot at:
https://github.com/nickg/nvc/issues/1004

nickg:
I don't really think it makes sense to have this as a global option as it now escalates assertion warnings to errors which overlaps with the existing `--exit-severity` and `--stderr` options. If it's needed at all it should be an analysis-only option.

Blebowski:
I will convert it to analysis option then.

Blebowski:
Now processed.

nickg:
This still has a problem where the option is left enabled after analysis completes.

nickg (approved):

Issue #1347, opened 2025-11-14
** Fatal: expression cannot be folded to an integer constant [case choice]


jevogel:
Constant record elements fail to be folded into integer constants in case choices.

Example `bug.vhd`:

```vhdl
entity bug is
end entity;
architecture test of bug is
  type reg_addrs_t is record
    A, B : integer;
  end record;
  constant REGS : reg_addrs_t := (A => 0, B => 1);
  signal sel, sig : integer;
begin
  process (all)
  begin
    case sel is
      when integer(REGS.A) => sig <= 0;
      when REGS.B => sig <= 1;
    end case;
  end process;
end architecture;
```

Running `nvc -a bug.vhd -e bug -r` produces the following output:

```text
** Fatal: expression cannot be folded to an integer constant
    > bug.vhd:14
    |
 14 |       when REGS.B => sig <= 1;
    |            ^^^^^^
```

Note that the issue can be resolved by casting the case choice to integer, but this is not ideal for my scenario where I have a lot of register addresses that I want to use this way in a register map.

PR #1354, opened 2025-11-28
Allow range of const record elements in case choice

NikLeberg:
I've extended our solution to #1347 with the capability to handle a range of constant record elements.
I don't think a slice i.e. association subkind `A_SLICE` is possible here. So I've left that out.
Cheers, Nik

### DIFF
--- a/src/simp.c
+++ b/src/simp.c
@@ -905,34 +905,11 @@ static bool simp_match_case_choice(tree_t alt, int64_t ival)
    return false;
 }
 
-static tree_t simp_case(tree_t t, simp_ctx_t *ctx)
+static tree_t simp_case(tree_t t)
 {
    const int nstmts = tree_stmts(t);
    if (nstmts == 0)
       return NULL;    // All choices are unreachable
-
-   for (int i = 0; i < nstmts; i++) {
-      tree_t alt = tree_stmt(t, i);
-      const int nchoices = tree_choices(alt);
-      for (int j = 0; j < nchoices; j++) {
-         tree_t a = tree_choice(alt, j);
-         if (tree_has_name(a)) {
-            tree_t n = tree_name(a);
-            switch (tree_kind(n)) {
-            case T_LITERAL:
-            case T_STRING:
-            case T_REF:
-               break;
-            default:
-               if (eval_possible(n, ctx->registry, ctx->mir)) {
-                  n = eval_try_fold(ctx->jit, n, ctx->registry, NULL, NULL);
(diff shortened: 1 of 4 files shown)


NikLeberg:
CI found another issue. I forgot to check if the tree node is already a `T_LITERAL`, `T_STRING` or `T_REF` and skip the eval if this is the case. I already pushed a fixup.

But now I'm not terribly happy about the function name:
```c
static tree_t simp_case_choice(tree_t t, simp_ctx_t *ctx);
```
Because it suggests that it handles the whole choice. But it only handles the name or left/right _subnode_. But naming wise in `simp.c` all functions handle parts of VHDL syntax and not tree specific nodes.
Maybe a new function in `eval.c` would be better?
```c
tree_t eval_try_fold_locally_static(tree_t expr, unit_registry_t *ur, mir_context_t *mc);
```

What would you prefer?

nickg:
> 
> But naming wise in `simp.c` all functions handle parts of VHDL syntax and not tree specific nodes. 

Well most are but there are some exceptions.

> 
> What would you prefer?

How about `simp_static_expr()` but otherwise keep it the same.

src/simp.c
```diff
@@ -927,20 +942,30 @@ static tree_t simp_case(tree_t t, simp_ctx_t *ctx)
...
+               assert(tree_ranges(a) == 1);
+               tree_t range = tree_range(a, 0);
+               tree_t l = tree_left(range);
+               tree_t lf = simp_case_choice(l, ctx);
+               if (l != lf)
+                  tree_set_left(range, lf);
+               tree_t r = tree_right(range);
+               tree_t rf = simp_case_choice(r, ctx);
+               if (r != rf)
+                  tree_set_right(range, rf);
```
nickg:
```suggestion
               tree_set_right(range, simp_case_choice(r, ctx));
```

(and elsewhere)

There's no harm in setting it to the same value and looks less noisy.

NikLeberg: I'd agree and that was actually how I initially did it. But with that I get several crashes of `Fatal: Write to object in frozen arena`. Is there something else I need to account for?

nickg: I guess that happens because `simp_case` is called from `simp_tree_global` as well `simp_tree_local` (I can't remember why) and the `T_ALTERNATIVE` won't necessarily have a writable copy during elaboration.

I think you could probably fix that by moving this logic to a new `simp_case_alternative()` that's called from `simp_tree_local` only for `T_ALTERNATIVE`.

Issue #1042, opened 2024-10-28
Random initialization


Blebowski:
A VHDL simulation initializes value of signal to `type'left`.
This may sometimes hide bugs on RTL, that later pop-out on GLS.

An example is FSM coded with enum type and first state being `S_IDLE`.
VHDL simulation initializes a state register to `S_IDLE`. If state register is not reset,
the RTL simulation will work correctly because FSM starts from correct state.
On GLS or on real device, there is no guarantee that the FSM state vector will initialize
to `S_IDLE` encoding.

This pops out in UPF flow simulations, where if a power domain with such FSM is disabled,
the FSM will immediately become `S_IDLE`. Missing reset after re-enabling the domain
will not be recognized.

What can be done on coding style level, is to define first state in enum type to something
like `S_UNDEFINED`. Such state will be default power-up state. Only upon reset, the state vector
will be initialized to `S_IDLE`. A state decoder will hide the `S_UNDEFINED` under `when others =>`
clause and either:
- fires an error if the FSM is to be launched and its state is `S_UNDEFINED`, or
- the transition from `when others` is encoded to go into `S_IDLE`.

This approach helps, but it does not solve the problem. `S_UNDEFINED` is just a proxy state
for RTL, that will be encoded into a state vector value by the synthesizer. So, the `when others`
clause may not execute in real device. In real device, a post power-up value of a state register
may be:
- any of valid FSM states.
- an invalid state depending on how FSM is encoded by synthesizer.

If it is valid state, the logic will behave as if the FSM was in that state. Likely, waiting for some
"done signal" it will stay stuck forever.

If it is an invalid state, it can be protected by the `when others => s_next <= S_IDLE` coding
approach.

This may sound like virtual issue, but I have seen it in reality and consequences were not nice.

On simulator level, what can be done, is to initialize signals randomly depending on some simulator
settings (.e.g `--init-signals=(lrm|random|0|1)`. I would propose following ways:

**LRM compliant:**
- As is now, the default behavior

**Random init**
- Enum types are initialized to a run-time decided random value. An exception is `std_logic` family of types that initializes to either `0` or `1`.
- Integer types initialize to a random value within its valid range.

**Fixed init (init_param = 0 or 1) **
- If `init_param=0`, all enum types are initialized to `type'left`. If `init_param=1`, all enum types are initialized to `type'right`. Exception is `std_logic` family of types that initialize to `0` or `1` based on `init_param`
[...]

PR #1384, opened 2026-01-18
Random signal initialization plugin.

Blebowski:
Hi,

this is a WIP proposal of random signal initialization VHPI plugin. Not yet ready to be merged, putting it here in case someone wants to have a look or provide comments.

The plugin provides random signal initialization options as described in: https://github.com/nickg/nvc/issues/1042
Configuration is done by `+siginit+` prefixed arguments.

The plugin build is hooked to the NVC build. I went for the simplest approach of custom target. Having __no__ experience with automake, google + AI helped me... hopefully not too bad :) Building native shared library required using LibTool extension, and I did not want to complicate it. Possible it will not work on Windows/Unix though.

The plugin load is integrated into `run_cmd`, plusargs are filtered, and only those with prefix `+siginit+` are passed to the plugin. This may be redundant (all plusargs could be passed). However, to pass seed, I extend the arguments with `+siginit+seed+<seed_value>`.

If `+siginit+seed+<seed_value>` is specified, it is used to randomize the signal init value, if not, the seed is taken from `--seed`.

The randomizer is "state-less", only signal name, index (if array is being initialized) and seed are used to generate random value.
This is to meet what was proposed in https://github.com/nickg/nvc/issues/1042 by Wilson Snyder.
Two simple and nested hashes are used as randomizer.

TODOs:
1. [x] Figure out the build on Windows / MAC
2. [x] Add record signal support -> Walk fields.
3. [x] Add some verbosity settings (e.g. if `+siginit+verbose` is set, value of every signal assigned is printed via `vhpi_printf`).
5. [x] Add support for specifing `entity` or hierarchy where this occurs. Currently it is whole design, but typically user would do this only on DUT, so either pass DUTs hierchy, or rather DUT entity name.
6. [x] Tests
7. [x] Documentation


nickg:
> The plugin build is hooked to the NVC build. I went for the simplest approach of custom target. Having no experience with automake, google + AI helped me... hopefully not too bad :) Building native shared library required using LibTool extension, and I did not want to complicate it. Possible it will not work on Windows/Unix though.

You can just copy what `test/vhpi/Makemodule.am` does. That already works on Windows and macOS.

Blebowski:
Hi @nickg,

this is now in first reasonably working state. I hand-tried on my project and NEORV32.

I have couple of questions before I proceed with writing Tests:
1. Are you OK with the `+siginit+` argument format as is now (functionality / format wise) ? 
2. I did not manage to solve the "arguments should not be passed to other VHPI plugins", is it acceptable ?
3. Is it sufficient to add couple of "regress" tests, or should this be completely new category (e.g. plugins) type of tests ?
4. When I try to do this on mixed language design, I get Fatal like: `Unsupported tree kind T_VERILOG in build_designUnit`. Can we filter and ignore Verilog units in the VHPI as long as there is no VPI support ?
5. I see strange behavior on records and arrays mutually nested. If I ommit `vhpi_get_str` with `vhpiNameP` when traversing record / array elements, I get weird string like `(NULL)(0)` then when I query the same property in the nested call on the same signal handle. Any hint on what might cause this is appreciated.
6. Are you OK with me having copyright on this ? It is in `contrib` and README states you don't accept another copyrights under the `src` folder. It will be still coupled to the main code-base via documentation and tests though. I would like to keep it close to the upstream, since maintaining the plugin separately would cause additional "compatibility" effort. I understand having copyright on this would bind me to some maintenance effort.

Blebowski:
Hi @nickg,

do you have any inputs to this please ?

nickg:
I added a `vhpiRandomSeedP` property to read the random seed in 1df46e2be9f491e43ef5.

Blebowski:
Now adapted to use the vhpiRandomSeedP.

Blebowski:
Finally managed to find time to debug the pipeline on this.
There were following issues:
- Install step needed to be added since the built-in plugin is loaded from "get_lib_dir()" during run time.
- Shared lib extension needed to be resolved for each platform separately. On windows, this needed to be distributed as "DATA" not "PROGRAMS" to avoid appending "exe" suffix.
- The pacman needed to have "override", I suspect because the files from manual install were conflicting with the ones from pacman install.

Blebowski:
Hi @nickg,

would you be open to looking into this please ?

From my side this is more-less complete. I was wondering if you are open to leaving me copyright on the plugin.
I understand this comes with maintenance responsibility (e.g. bug occurs, or you ping me with something related to it).
I am not pushing for it, I can give the copyright to you. I only would like to avoid maintaining this "out of mainline" as it will likely die soon then.

nickg:
I would prefer it to work like `--load siginit` and not rely on special argument parsing. I added a search path in fcb4ee747bc which can be used for this and also if `run_regr` sets `NVC_PLUGIN_PATH` to the `lib/` directory in the build tree you don't need to change GHA workflows. 

> I was wondering if you are open to leaving me copyright on the plugin.

I am OK with this but only for code outside of `src/`.

Blebowski:
Thanks.

I adapted the implementation to kick-in with the `--load=siginit` and used the `NVC_PLUGIN_PATH` in regression.

There are still few corner-cases (e.g. with 2019, I don't think integers will work correctly as vhpiIntT is 32 bit only),
but that can be iterated later.

src/jit/jit-ffi.c
```diff
@@ -147,7 +147,7 @@ jit_dll_t *ffi_load_dll(const char *path)
    if (path == NULL)
       return ffi_load_exe();
 
-   char *abs = realpath(path, NULL);
+   char *abs = realpath_if_exists(path);
```
nickg: I fixed this in dbf4b7801bff6f

src/util.h
```diff
@@ -85,6 +85,16 @@
...
+char *realpath_if_exists(const char *path);
+
+#ifdef __MINGW32__
+#define DLL_EXT "dll"
+#elif defined(__APPLE__)
+#define DLL_EXT "dylib"
+#else
+#define DLL_EXT "so"
+#endif
+
```
nickg: This is already set in `config.h`

contrib/plugins/Makemodule.am
```diff
@@ -0,0 +1,33 @@
...
+all-local: lib/nvc/siginit.dll
+
+nodist_proxy_DATA = lib/nvc/siginit.dll
+
+CLEANFILES += lib/nvc/siginit.dll contrib/plugins/siginit.o
+
+else
+if MACOS_BUILD
+proxy_PROGRAMS = lib/siginit.dylib
+lib_siginit_dylib_SOURCES = contrib/plugins/siginit.c
```
nickg: You can interpolate the name like `lib_siginit_@DLL_EXT@_SOURCES`. There's no need to duplicate the rules for each platform.

Blebowski: I deduplicated this for Linux and MACOS.

For windows, I was fighting with automake adding `.exe` suffix to the built library when I used `_PROGRAMS`.
Other variant was to use "Libtool, which I wanted to avoid.

Sorry for the mess, I am much more used to CMake, so I am still rather fighting the automake than embracing it.

Blebowski:
Any more comments on this MR please ?

contrib/plugins/Makemodule.am
Blebowski: I rewrote the build setup to be custom make rule for all platforms. This way it is unified and does not rely on _PROGRAMS anymore. Only custom handling is the IMPLIB_REQUIRED.

Blebowski:
Hi @nickg,

do you think these changes can be merged please ? Or is there something else to fix in this ?

nickg:
I fixed some things with the makefile. I'm not sure there's much point having a separate `testlist.txt`, it can be merged into the main one.

test/run_regr.c
```diff
@@ -138,7 +139,9 @@ struct test {
    char      *cover;
    char      *define;
    char      *export;
-   char      *plusarg;
+   char     **plusargs;
+   unsigned   nplusargs;
```
nickg: I added multiple plusargs in 5e1c48dc9

contrib/plugins/siginit.c
```diff
@@ -0,0 +1,614 @@
...
+    cb.reason = vhpiCbStartOfSimulation;
+    cb.cb_rtn = vhpi_cb;
+    cb.user_data = (vhpiHandleT)&ctx;
+
+    vhpi_register_cb(&cb, 0);
+
+    vhpiErrorInfoT error_info;
+    vhpi_check_error (&error_info);
+
+    if (error_info.severity > 0) {
```
nickg: You need to check the `vhpi_check_error` return code. The `error_info` struct is not initialised if there is no error.

test/run_regr.c
Blebowski: I removed the original approach then and fixed minor thing causing plusargs passed as global arguments not to be passed to the plugin.

Blebowski:
> I fixed some things with the makefile. I'm not sure there's much point having a separate testlist.txt, it can be merged into the main one.

Where can I find them ? I did not find any branch in your repo. Is it possible you pushed them to the branch in my fork and I overwrote them by force-push after rebase without pulling first ?

Blebowski:
I have rebased to the latest master, and merged the regress test lists.

nickg:
> 
> Where can I find them ? I did not find any branch in your repo. Is it possible you pushed them to the branch in my fork and I overwrote them by force-push after rebase without pulling first ?

I guess I just forgot to push it.

Blebowski:
Hi @nickg , are you open to reviewing and merging this please ?
If not, I can close this MR, and move it somewhere else (e.g. a stand-alone repo, possibly with multiple plugins).
However, I would like to keep it close to mainline since maintaining and releasing this separately has unnecessary
effort in testing. I hope this is useful feature, and I will try to do my best maintaining the plugin.

nickg (approved):

nickg:
FWIW it should be possible to do this kind of thing with TCL scripts in future which has better error handling and automatic releasing of VHPI handles. See `test/regress/vhpi20.tcl` for an example.

Blebowski:
Thanks, noted. For this particular plugin, it is definitely a good option. Do you know if TCL is in general comparable to C when e.g. dealing with large (or lot of small) objects ? 

E.g. the SDF parser + annotator can be implemented as VHPI plugin (and potentially in TCL). I would guess it would have
worse performance than having native C implementation, and tight integration with elaborator (rather than milions of VHPI calls and TCL on top of it).

Issue #1406, opened 2026-02-03
BOOLEAN_VECTOR arrays not dumped correctly in FST waveforms


jonahshader:
Minimal test case:

```vhdl
library ieee;
use ieee.std_logic_1164.all;

entity boolean_array_bug is
end entity;

architecture test of boolean_array_bug is
    type state_t is (IDLE, RUNNING, DONE);
    type state_array_t is array (0 to 1) of state_t;
    signal state_arr : state_array_t := (others => IDLE);

    type bool_array_t is array (0 to 1) of BOOLEAN;
    signal bool_arr : bool_array_t := (others => false);

    signal bool_vec : BOOLEAN_VECTOR(0 to 1) := (others => false);
begin
    process begin
        wait for 1 ns;
        state_arr(0) <= RUNNING;
        bool_arr(0) <= true;
        bool_vec(0) <= true;

        wait for 1 ns;
        state_arr(1) <= RUNNING;
        bool_arr(1) <= true;
        bool_vec(1) <= true;
        wait for 1 ns;
        wait;
    end process;
end architecture;
```

Run with `nvc -a test.vhd -e boolean_array_bug -r --wave=test.fst --dump-arrays`

We expect all three arrays to show individual elements, but instead we only see individual elements for `state_arr`, with `bool_arr` and `bool_vec` only showing single values.

<img width="1157" height="179" alt="Image" src="https://github.com/user-attachments/assets/f6b4d23f-50d9-41a0-aab4-98aa4bdfc5bc" />

The fix in 8ac16bc for #1362 handles custom enum arrays but not boolean arrays. The condition `ft->sdt != FST_SDT_NONE` is true for boolean, so boolean arrays still take the collapsed single-handle path. I've got a fix for this ready, I'll open a PR shortly.

PR #1407, opened 2026-02-03
Fix waveform dumping for BOOLEAN_VECTOR

jonahshader:
A fix for issue #1406

What else does this PR need to get added? Do you want me to add a regression test, similar to how 8ac16bc does it?


nickg:
> What else does this PR need to get added? Do you want me to add a regression test, similar to how [8ac16bc](https://github.com/nickg/nvc/commit/8ac16bc83d77c2cf0a5cec3665ebae62690668c7) does it?

Yes that would be good. You can do `make bin/fstdump` and then `./bin/fstdump out.fst` to generate the comparison text file.

PR #1447, opened 2026-03-08
Adding VHDL-2019 range records.

### LINKED ISSUE #1390: VHDL-2019: Extended Ranges
I know the feature is not implemented, but I wanted to try and post a use case and an example here.

The idea that we have a full range of integers, but only a small set are in the set. This is a way to set up discrete sets. Maybe useful for coverage.

```vhdl
entity test is
end entity ;

architecture arch of test is

    type ranges_t is array(natural range <>) of integer'range'record ;

    subtype a is integer range -10 to 0 ;
    subtype b is integer range 100 to 200 ;
    subtype c is integer range 4929 to 8388 ;

    constant set : ranges_t := (a'range'value, b'range'value, c'range'value) ;

    function in_range(x : in integer ; set : ranges_t) return boolean is
    begin
        for idx in set'range loop
            if set(idx).direction = ASCENDING then
                if x >= set(idx).left and x <= set(idx).right then
                    return true ;
                end if ;
            else
                if x >= set(idx).right and x <= set(idx).left then
                    return true ;
                end if ;
            end if ;
        end loop ;
        return false ;
    end function ;

begin

    tb : process
    begin
        report to_string(in_range(100, set)) ;
        std.env.stop ;
    end process ;

end architecture ;
```

I don't necessarily know if I can get the range directly in the constant, so I added the subtypes.

Current error is:
```
$ nvc --std=2019 -a sets.vhdl 
** Error: unexpected record while parsing attribute name, expecting one of range, reverse_range, identifier or subtype
   > sets.vhdl:6
   |
 6 |     type ranges_t is array(natural range <>) of integer'range'record ;
   |                                                               ^^^^^^ this token was unexpected
** Error: prefix of user defined attribute reference cannot denote a sub-element or slice of an object
   > sets.vhdl:6
   |
 6 |     type ranges_t is array(natural range <>) of integer'range'record ;
   |                                                 ^^^^^^^^^^^^^
** Error: range expression not allowed here
    > sets.vhdl:12
    |
 12 |     constant set : ranges_t := (a'range'value, b'range'value, c'range'value) ;
```

bpadalino:
Attempt to add range records ([LCS2016_099](http://www.eda-twiki.org/cgi-bin/view.cgi/P1076/LCS2016_099)).

I had to add the definition of the `RANGE_DIRECTION` type in [standard.vhd](./lib/std.19/standard.vhd) so `BOOLEAN` could be seen by the predefined operators.

Note in the [official standard.vhd](https://gitlab.com/IEEE-P1076/packages/-/blob/develop/std/standard.vhdl?ref_type=heads#L41) it's called `DIRECTION` but in the LRM it is shown as `RANGE_DIRECTION`. Some changes might be needed if my reading was incorrect.

I used the `bootstrapping` flag to defer creation of the range records for `STANDARD`. Otherwise, they should be created right after the type is declared.

Note that this fixes #1390.

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -1252,27 +1252,26 @@ static void declare_alias(tree_t container, tree_t to, ident_t name)
    tree_add_decl(container, alias);
 }
 
-static void declare_range_record_type(tree_t container, type_t scalar_type)
+static void declare_range_record_type(tree_t container, tree_t td)
 {
-   type_t base = type_base_recur(scalar_type);
+   type_t type = tree_type(td);
    type_t dir_type = std_type(NULL, STD_RANGE_DIRECTION);
 
-   LOCAL_TEXT_BUF tb = tb_new();
-   tb_istr(tb, type_ident(base));
-   tb_cat(tb, "_range_record");
+   ident_t name = ident_sprintf("%s_range_record", istr(tree_ident(td)));
 
    type_t rec = type_new(T_RECORD);
-   type_set_ident(rec, ident_new(tb_get(tb)));
+   type_set_ident(rec, name);
+   mangle_type(nametab, rec);
 
    tree_t f_left = tree_new(T_FIELD_DECL);
    tree_set_ident(f_left, ident_new("LEFT"));
-   tree_set_type(f_left, base);
+   tree_set_type(f_left, type);
    tree_set_pos(f_left, 0);
    type_add_field(rec, f_left);
 
    tree_t f_right = tree_new(T_FIELD_DECL);
    tree_set_ident(f_right, ident_new("RIGHT"));
-   tree_set_type(f_right, base);
+   tree_set_type(f_right, type);
    tree_set_pos(f_right, 1);
    type_add_field(rec, f_right);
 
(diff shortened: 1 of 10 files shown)


src/parse.c
```diff
@@ -2132,6 +2132,27 @@ static type_t apply_index_attribute(tree_t aref)
    return sub;
 }
 
+static void ensure_range_record_ops(type_t rec, const loc_t *loc)
```
nickg: Do these predefined records types even have predefined operations? It's not clear to me from the LRM that they necessarily do.

bpadalino: Latest version keeps `=`, '/=`, and `to_string` which are defaults for all other record types. No other types are added.

nickg:
> I used the bootstrapping flag to defer creation of the range records for STANDARD. Otherwise, they should be created right after the type is declared.

This needs reporting on the IEEE issue tracker. The previous standards took care to ensure the `standard` package could be analysed as written.

bpadalino:
I've added the issue [Standard Package Circular Reference between `RANGE_DIRECTION` and `BOOLEAN`](https://gitlab.com/IEEE-P1076/VHDL-Issues/-/work_items/378) to the LRM Issues over at gitlab.

PR #1449, opened 2026-03-10
Add JIT intrinsics for more pkg_textio functions

malles030:


nickg:
Rather than just pasting the output from an LLM it would be helpful if you provided a test case that shows the problem you are trying to solve.

malles030:
Nick, please apologise. It's the first time I'm trying to make a contribution to an open source project. I will try to do better in the future.

I'm using a testbench that relies heavily on File-IO for reading GBs of test vectors. For that I increased the heap to a high value like 12 GB. Currently, the heap grows by two entries per line in the text file, as VHDL readline creates two memories: First, the initial memory (1 to 127), second the memory shrinked to the actual size of the line.

Since my heap is so big (as I want so save some GBs of data), the heap is not cleared by the garbage collection, if I understand the mechanism correctly. Only when the heap is full, garbage collection kicks in, but in my simulation it then kind of looks like the simulation is stuck while loading the files as it has to look through GBs of data.

I think the core of the problem is that I have objects on the heap that are big and supposed to live throughout the simulation as well as millions of tiny allocations caused by millions of readline calls.

This commit does not resolve this issue but it mitigates it as the heap grows only half as fast with the intrinsic (one allocation per line only). Furthermore, the intrinsic runs much faster. Here are the results of the test-case that I attached:

Without the change:
```
** Note: setup:37ms run:4594ms user:4543ms sys:175ms maxrss:864492kB static:0kB
```

With the change:
```
** Note: setup:35ms run:1675ms user:1665ms sys:93ms maxrss:487824kB static:0kB
```

Run-time reduction by 2.7x, maxrss reduction by 1.8x


To run I used
```
./bin/nvc -L lib/ -a ../test/regress/textio_mem_tb.vhd -e -gG_NUM_LINES=500000 textio_mem_tb
./bin/nvc -L lib/ -H 4G  -r --stats textio_mem_tb
```

I hope that helps.


[textio_mem_tb.vhd.zip](https://github.com/user-attachments/files/25897480/textio_mem_tb.vhd.zip)

Blebowski:
@nickg seeing this, I was wondering if you have something like "AI policy" in mind for contributors ? I hear about OS maintainers get swamped by dozens of AI generated MRs, and I don't want it to happen in this project.

nickg:
> Currently, the heap grows by two entries per line in the text file, as VHDL readline creates two memories: First, the initial memory (1 to 127), second the memory shrinked to the actual size of the line.
> [..]
> This commit does not resolve this issue but it mitigates it as the heap grows only half as fast with the intrinsic (one allocation per line only).

I don't think this is true though: the `shrink` call gets subbed out for an intrinsic which adjusts the bounds of the existing allocation (right above the code you added!). 

The main problem is `readline` calls `read` for each character which involves a relatively expensive VHDL->runtime transition which you avoid here by intrinsifying the whole procedure. An easy win is to remove the `endfile` call as the check for the number of characters read handles that anyway. I did that in 183b993 and it drops the time from ~24s to ~17s on my machine.

> Since my heap is so big (as I want so save some GBs of data), the heap is not cleared by the garbage collection, if I understand the mechanism correctly. Only when the heap is full, garbage collection kicks in, but in my simulation it then kind of looks like the simulation is stuck while loading the files as it has to look through GBs of data.

This has been reported before and admittedly the current design is not great. It would be better to incrementally enlarge the heap up to the limit given by the user based on some heuristics like occupancy and time between collections.

> I think the core of the problem is that I have objects on the heap that are big and supposed to live throughout the simulation as well as millions of tiny allocations caused by millions of readline calls.

In general time taken by GC time is proportional to the number of _live_ objects so as long as you don't hold onto references to those small allocations they are essentially invisible when tracing the heap. Unfortunately in NVC GC time is also proportional to the size of live objects so that might be causing an issue here. You can try setting `NVC_GC_VERBOSE=1` and that will log each GC cycle and the total time taken. On the test case above it only accounts for 0.4% of the total runtime though.

nickg:
> @nickg seeing this, I was wondering if you have something like "AI policy" in mind for contributors ? I hear about OS maintainers get swamped by dozens of AI generated MRs, and I don't want it to happen in this project.

I don't have a moral problem it: they are tools and have uses. For example I used Codex to write some of the Verilog bignum maths routines. But there needs to be a human involved at some level checking the output, and as you say there is an asymmetry problem where it's easy to generate huge volumes of code which some human maintainer then has to go and read and be responsible for indefinitely.

malles030:
> In general time taken by GC time is proportional to the number of _live_ objects so as long as you don't hold onto references to those small allocations they are essentially invisible when tracing the heap. Unfortunately in NVC GC time is also proportional to the size of live objects so that might be causing an issue here. You can try setting `NVC_GC_VERBOSE=1` and that will log each GC cycle and the total time taken. On the test case above it only accounts for 0.4% of the total runtime though.

I dug deeper into the issue as it turned out that the GC does its job correctly. It found out that the allocation of small chunks of data (51 bytes) used forever (mspace_try_alloc) since at some point there where millions of 32 byte fragments in the free_list. The cause turned out to be a scheme of "a small persistent allocation -> readline -> small persistent allocation -> readline etc":

```
for i in 0 to v_num_lines - 1 loop
    v_data_sets_new.all(v_num_files).data.all(i) := new t_int_array(0 to columns - 1);
    readline(file_handler, v_line_in);
    -- fill v_data_sets_new.all(v_num_files).data.all(i).all(j) with column data here
end loop;
```
The following change in my code resolves the hanging of NVC:
```
for i in 0 to v_num_lines - 1 loop
    v_data_sets_new.all(v_num_files).data.all(i) := new t_int_array(0 to columns - 1);
end loop;
for i in 0 to v_num_lines - 1 loop
    readline(file_handler, v_line_in);
    -- fill v_data_sets_new.all(v_num_files).data.all(i).all(j) with column data here
end loop;
```
Not sure whether there is something one can do about in NVC easily. The code works nicely in three other simulators.

So the outcome for me is that this PR only helps to improve NVC's performance and that it reduces simulation startup time by about 20s for me. Reading from files is still quite a bit slower than commercial tools but the gap is much smaller.
I also tried your quick fix, it also helps to speed up the readline call, but not as much as the intrinsics.

PR #1451, opened 2026-03-10
Add missing V_FOR_LOOP in build_sensitivity for Verilog always blocks

malles030:


malles030:
Just to add some background: I tried to compile a real-world Verilog design, which is using arrays extensively. I got this error message:

```
cannot handle V_FOR_LOOP in build_sensitivity
```

as some signals from the sensitivity list are accessed from within a for loop. This PR resolves this issue for me.

PR #1452, opened 2026-03-11
Implement fast paths for mul JIT intrinsics

malles030:
The functions of numeric_std run faster if we implement fast paths for computations if the results are less than or equal to 64 bits.

An FIR filter in a typical DSP scenario (100 taps, 25 and 18 bits operands for multiplication) runs 3.2x faster with the changes, most of it (3.1x) stemming from the multiplication fast path, avoiding the approach to realize the multiplication with additions only.

With packing to 64-bits, a plain simulation with 32-bit additions runs 13% faster, 48-bit additions still give a 7% gain. With 6 bits operands, addition speed remains the same.

### DIFF
--- a/src/jit/jit-intrin.c
+++ b/src/jit/jit-intrin.c
@@ -541,6 +541,33 @@ static inline void __spread_bits_8(void *vec, uint8_t packed)
    unaligned_store(vec + 4, spread_nibble[(packed >> 0) & 0xf], uint32_t);
 }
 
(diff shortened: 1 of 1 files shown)


malles030:
I append the filter-like VHDL code I used for benchmarking the changes.

[fir_filter.zip](https://github.com/user-attachments/files/25903157/fir_filter.zip)

PR #1456, opened 2026-03-16
Collect code coverage for package declarations

### LINKED ISSUE #1007: Code coverage data in packages are missing
Currently code coverage is not reported for packages.

Similar to:
[Packages missing in waves](https://github.com/nickg/nvc/issues/901)

malles030:
Previously, coverage instrumentation was only applied to design hierarchy instances.  Top-level packages loaded via USE clauses were lowered without coverage data, so any functions, procedures, or expressions defined in packages were invisible to the coverage report.

Pass coverage data through the unit registry so that lower_package and lower_pack_body can create coverage scopes under the root scope.  A new helper vhdl_cover_package walks the package and body declarations to register coverage items.  Standard libraries (STD, IEEE, NVC) are excluded.

Fixes #1007

### DIFF
--- a/src/cov/cov-data.c
+++ b/src/cov/cov-data.c
@@ -876,6 +876,11 @@ bool cover_enabled(cover_data_t *data, cover_mask_t mask)
    return data != NULL && (data->mask & mask);
 }
 
+cover_scope_t *cover_get_root_scope(cover_data_t *db)
+{
+   return db != NULL ? db->root_scope : NULL;
+}
+
 cover_scope_t *cover_create_block(cover_data_t *db, ident_t qual,
                                   cover_scope_t *parent, tree_t inst,
                                   tree_t unit)
@@ -933,7 +938,7 @@ cover_scope_t *cover_create_scope(cover_data_t *db, cover_scope_t *parent,
 
    assert(parent != NULL);
    assert(db->root_scope != NULL);
-   assert(!is_design_unit(t));
+   assert(!is_design_unit(t) || is_package(t));
 
    cover_scope_t *s = pool_calloc(db->pool, sizeof(cover_scope_t));
 
@@ -971,7 +976,16 @@ cover_scope_t *cover_create_scope(cover_data_t *db, cover_scope_t *parent,
    s->block  = parent->block;
    s->name   = name;
    s->loc    = *tree_loc(t);
-   s->hier   = ident_prefix(parent->hier, s->name, '.');
+
+   // Top-level packages already carry their library in tree_ident
+   // (e.g. LIB1.PKG1). Storing the hierarchy as
+   // <work-lib>.<pkg-lib>.<pkg-name> would be misleading, so use the
+   // package ident directly as the hier path.
+   if (s->kind == CSCOPE_PACKAGE && parent == db->root_scope)
+      s->hier = s->name;
+   else
+      s->hier = ident_prefix(parent->hier, s->name, '.');
+
    s->emit   = cover_should_emit_scope(db, s);
 
    if (s->sig_pos == 0)
--- a/src/cov/cov-exclude.c
+++ b/src/cov/cov-exclude.c
@@ -17,6 +17,7 @@
 
 #include "util.h"
 #include "array.h"
+#include "common.h"
 #include "cov/cov-api.h"
 #include "cov/cov-data.h"
 #include "ident.h"
@@ -29,11 +30,21 @@
 
 typedef struct _excl_children excl_children_t;
 typedef struct _excl_trie excl_trie_t;
+typedef struct _excl_glob excl_glob_t;
+
+struct _excl_glob {
+   char       *glob;
+   bool        include;
+   excl_glob_t *next;
+};
 
 typedef struct _cover_spec {
    excl_trie_t *hier_trie;
+   // TODO: convert block_trie and fsm_trie to glob lists so users can write
+   // e.g. "+fsm_type *_fsm_t" matching across name components.
    excl_trie_t *block_trie;
    excl_trie_t *fsm_trie;
+   excl_glob_t *pkg_globs;
 } cover_spec_t;
 
 typedef enum {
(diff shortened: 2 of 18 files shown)


malles030:
@nickg This  is up to discussion of course. Personally, I'm not super happy with the approach to consider each package for coverage except for the filtered ones. I'd prefer to have a new 'file' keyword in the coverage specification file that would allow to enable coverage for the given files only. This is typically supported by commercial tools and allows for a simple scripted generation of the coverage specification file (without having to think about the hierarchy), taking all synthesisable code into account but omitting testbench files.
Maybe by default one could collect coverage data for packages in the work library only. And for more complex designs, which are using multiple VHDL libraries, it  then becomes necessary to setup the coverage specification file. What do you think?

Blebowski:
Hi,

my small bits to this:
- since there exists `hier` in the coverage specification file, the package could be specified like: `hier LIBRARY_NAME.PACKAGE_NAME`. 
- Alternatively, there could be `package *` to collect all packages or `package <PACKAGE_NAME>` to collect only single package.

Blebowski:
Hi,

I was testing this MR, and there is issue that the hierarchy path of the packages gets the design top library prepended, e.g.:

image

image

I think this needs to be fixed.

As for the specification of what packages to collect, I propose:
- By default no packages are collected
- If package hierarchy path is matched by `hier` in coverage spec, it is collected.
- If package name is matched by `+package <pattern>` it is collected. Using `-package <pattern>` excludes package matching the pattern. So e.g.:
```
+package *_PKG
-hier VUNIT_LIB*
```
would collect all packages whose name is suffixed with `_pkg`, except those compiled in `vunit_lib`.

malles030:
@Blebowski I took your suggestions into account and extended the documentation accordingly. Any further feedback is welcome. Selecting the coverage scope with a +file parameter probably fits better to a separate PR.

malles030:
Just as an addition:

- HTML nav-tree now groups top-level scopes by their own library: each library renders as its own collapsible details block and the root anchor is labelled root instead of the work-library name.
- Inside a library group the entries display the bare scope name (e.g. MATH_PKG) so the library prefix isn't repeated, and only the library matching the currently viewed scope is expanded by default.



image

src/cov/cov-data.c
```diff
@@ -933,7 +938,8 @@ cover_scope_t *cover_create_scope(cover_data_t *db, cover_scope_t *parent,
 
    assert(parent != NULL);
    assert(db->root_scope != NULL);
-   assert(!is_design_unit(t));
+   assert(!is_design_unit(t) || tree_kind(t) == T_PACKAGE
```
Blebowski: There exists `is_package` function for this purpose.

src/lower.c
```diff
@@ -13033,11 +13076,13 @@ vcode_unit_t unit_registry_get(unit_registry_t *ur, ident_t ident)
             return NULL;
 
          unit_registry_defer(ur, unit_name, NULL, emit_package,
-                             lower_pack_body, NULL, tree_to_object(body));
```
Blebowski: Now that `ur` contains `cover` directly, could the `unit_registry_defer` be refactored not to pass the `cover` explicitly anymore, but instead always take `ur->cover` ? AFAICT, there is always only single `cover` struct during one elab/run.

src/cov/cov-exclude.c
```diff
@@ -29,11 +30,19 @@
...
+   char       *glob;
+   bool        include;
+   pkg_glob_t *next;
+};
 
 typedef struct _cover_spec {
    excl_trie_t *hier_trie;
    excl_trie_t *block_trie;
    excl_trie_t *fsm_trie;
+   pkg_glob_t *pkg_globs;
```
Blebowski: I suggest to rename this to sth like `excl_glob_t`. 

The "trie" approach fits for fast search over comma separated hierarchy with wild-card at the end to denote the subscope.

The flat list of globs would be more fitted also in case of `block_trie` and `fsm_trie`. Users could e.g. choose strategy to name all FSM enum types with `fsm_t` suffix. Enabling FSM coverage collection in such case would be as simple as `+fsm *_fsm_t`.

If @nickg agrees, please leave a comment above "block/fsm" trie with sth like "TODO: convert block and fsm tries to globs".

src/cov/cov-exclude.c
```diff
@@ -553,6 +596,55 @@ void cover_load_spec_file(cover_data_t *db, const char *path)
...
+}
+
+bool cover_should_emit_package(cover_data_t *db, tree_t pack)
+{
+   // Packages are not collected by default; the user must opt-in via a
+   // coverage spec file with +package or +hierarchy directives.
+   if (db == NULL || db->spec == NULL)
+      return false;
+
+   // Standard libraries are never instrumented automatically. The user
```
Blebowski: Does there need to be specific rule for the built-in packages ?

If this special section was completely dropped, wouldn't the `cover_match_packages` still
work with stuff like `+hier ieee.numeric_std` or with `+package numeric_std` ? 

Even if so, I can't figure out use-case where someone would like to collect code coverage
for language built-in packages. Plus, some of the IEEE functions are replaced by intrinsics
for speed, so the coverage numbers would not be exact anyway. My personal preference
would be to always skip these, and return false here.

Blebowski (commented):
Thanks @malles030, I tested this and it works really nicely.

I attached couple of comments after reviewing this.

src/cov/cov-data.c
malles030: Done.

src/lower.c
malles030: Yes, that works. It's within the latest push.

src/cov/cov-exclude.c
malles030: Also done.

src/cov/cov-exclude.c
malles030: OK, they are removed now and it's not possible anymore to add them explicitly.

obruendl:
@malles030 As you requested in #1525 I tested this branch here regarding package coverage.

I did NOT find coverage for packages being reported. The TBs for packages I see (e.g. olo_base_pkg_math_tb) but the package itself (e.g. olo_base_pkg_math) I do not see.

I use the coverage reported through stdout or stderr currently. I have attached the output. Could it be that simply the output is missing?

[nvc_coverage.txt](https://github.com/user-attachments/files/28473260/nvc_coverage.txt)

malles030:
@obruendl Thanks for testing. Can you confirm that you added a line like
```
+package *pkg*
```
to your cover-spec file?

malles030:
Closing, no feedback from author and superseded by https://github.com/nickg/nvc/pull/1634
Good luck @Blebowski! :-)

PR #1458, opened 2026-03-17
Fix replication operator with braced constant expression in Verilog

malles030:
The Verilog parser incorrectly rejected replication operators where the repeat count was a constant expression enclosed in braces, such as {{N+1}{1'b0}}. The inner {N+1} was parsed as a single-element concatenation rather than being unwrapped to the expression N+1, causing the semantic check to report "expression is not a constant".

### DIFF
--- a/src/vlog/vlog-parse.c
+++ b/src/vlog/vlog-parse.c
@@ -1598,6 +1598,12 @@ static vlog_node_t p_multiple_concatenation(vlog_node_t head)
 
    BEGIN_WITH_HEAD("multiple concatenation", head);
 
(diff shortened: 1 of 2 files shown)


nickg:
>  The inner {N+1} was parsed as a single-element concatenation rather than being unwrapped to the expression N+1, causing the semantic check to report "expression is not a constant".

But it _is_ a single element concatenation. The repetition count for a multiple concatenation is an arbitrary expression which could be another concatenation. The problem is that concatenations are not recognised as constant expressions when all the elements are constant. I fixed that in f2c305f934f78.

malles030:
@nickg Great, thanks a lot.

PR #1467, opened 2026-03-27
Fix generic array override parsing for VHDL-2008

### LINKED ISSUE #1465: Fail to parse array values for generics in VHDL 2008
Hi, I recently came across this problem. 

I have a module which defines a generic of type unconstrained array:
```vhd
entity my_entity is
    generic (
        ALLOWED_INPUT_ITEMS         : generic_positive_array
    );
    port (
        clk_i : in std_logic
    );
end;
```

Where `generic_positive_array` is:
```vhd
type generic_positive_array is array(natural range <>) of positive;
```

When I try to analyze the entity I specify the generic through the argument `-gALLOWED_INPUT_ITEMS=(1,2,3,4,8,6)` nvc fails with this error message:

```txt
 Fatal: (init): failed to parse "(1,2,3,4,8,6)" as type GENERIC_POSITIVE_ARRAY for generic ALLOWED_INPUT_ITEMS
```

This is a functionality that some commercial simulators have, like riviera pro, and it would be great if nvc will be able to implement the same functionality!

rameloni:
Fixes #1465. Add support for parsing command-line generic overrides for unconstrained array types during elaboration.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 ## Unreleased changes
 - Added basic support for System Verilog assignment pattern expressions.
+- Generics of an array type other than a character array can now be
+  overridden on the command line with an aggregate value, for example
+  `-gVALUES='(1, 2, 3)'` (#1465).
 
 ## Version 1.22.1 - 2026-08-01
 - Waveform dumping now dumps multiple top units instead of only the first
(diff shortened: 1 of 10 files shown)


rameloni:
@nickg this feature is supported in some commercial simulators, like Riviera PRO. It would be nice to support this in nvc as well.

PR #1486, opened 2026-04-09
Verilog: fix wire initialiser scoping inside generate blocks

gitmodimo:
Wire declarations with initialisers inside generate blocks (IEEE 1800-2017 §27.3) fail at elaboration with "missing variable". simp_net_decl expands wire x = val into wire x + assign x = val at the module scope, but generate blocks form their own scope — the assign ends up in the wrong scope and loses pointer identity after vlog_copy.

Fix: perform the expansion as a post-processing walk in vlog_simp after vlog_rewrite completes (when if-generates are already resolved into V_BLOCK bodies). The new
expand_net_initialisers function recurses through sub-scopes, adding each continuous assign to the correct enclosing scope.

- Replace simp_net_decl wire initialiser expansion with expand_net_initialisers post-pass in vlog_simp                                                                                     
- Add assert(!vlog_has_value(v)) in vlog_lower_net_decl to catch any un-expanded initialisers                                                                                              
- Add regression test vlog38

### DIFF
--- a/src/vlog/vlog-simp.c
+++ b/src/vlog/vlog-simp.c
@@ -60,27 +60,6 @@ static vlog_node_t simp_net_decl(vlog_node_t decl, vlog_node_t mod)
       vlog_add_stmt(mod, g);
    }
 
-   if (vlog_has_value(decl)) {
-      vlog_node_t value = vlog_value(decl);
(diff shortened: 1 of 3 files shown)


nickg:
I fixed this in a different way in dd9fdd4c404

PR #1509, opened 2026-04-20
Verilog: Fix ? don't-care digit in casez/casex and number literals     

gitmodimo:
- casez/casex comparison was broken when don't-care (?/z) appeared in case items against a plain expression — the JIT irgen used logical NOT  (!) instead of bitwise NOT (~) for z-mask computation, making every casez degenerate to exact match. Existing tests only had z in both the  expression and item so they passed by coincidence.                                                                                            
  - Lexer couldn't tokenise ? followed by _ in number literals (e.g. 8'b1??z_zz??) — added VLOG_DIGIT class that includes ? alongside _.        
  - JIT wide EXTRACT assert on >64-bit operands — added JIT_VEC_EXTRACT exit handler for both narrow and wide results.                          
  - VPI vpi_get_value assert on >64-bit operands — read args as pointers for wide values.

### DIFF
--- a/src/jit/jit-exits.c
+++ b/src/jit/jit-exits.c
@@ -1105,6 +1105,66 @@ void __nvc_vec4op(jit_vec_op_t op, jit_anchor_t *anchor, jit_scalar_t *args,
          args[1].pointer = bresult;
       }
       break;
+   case JIT_VEC_EXTRACT:
+      {
+         assert(size > 64);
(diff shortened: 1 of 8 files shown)


nickg:
I pushed the lexer fix and test cases in 3ac3748d5af705ee4c4cc0c69853235d52024577

PR #1513, opened 2026-04-20
Verilog: Support default values on macro functions

Anselmo95:
Hello nickg,

The preprocessor has been extended to recognise [], {} and = as  tokens. The first two were needed for 'expression' parsing (the macro actual arguments and defaults) and the second to recognise if a macro argument has a default.

Added a list of text buffers for the defaults value and an utility function to check if a text buffer is empty or contains only spaces.

The p_expression function has been reworked to parse matching tokens (curly and square braces).
The rest of the logic is the parsing of default values and using them when no argument is provided to a macro.

### DIFF
--- a/src/vlog/vlog-pp.c
+++ b/src/vlog/vlog-pp.c
@@ -41,12 +41,14 @@ struct _ifdef_stack {
 };
 
 typedef A(ident_t) arg_list_t;
+typedef A(text_buf_t *) def_list_t;
 
 typedef struct {
    ident_t     name;
    loc_t       loc;
    text_buf_t *text;
    arg_list_t  args;
+   def_list_t  defs;
    bool        has_args;
    bool        error;
 } macro_t;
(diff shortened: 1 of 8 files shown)


src/vlog/vlog-pp.c
```diff
@@ -389,6 +404,61 @@ static token_t vlogpp_lex(void)
    }
 }
 
+static void p_expression(text_buf_t *tb, token_t end, bool stop_at_comma)
```
Anselmo95, line 407: I don't particularly like the last boolean argument but I could not come up with anything better

src/vlog/vlog-pp.c
```diff
@@ -552,11 +607,24 @@ static void p_list_of_actual_arguments(macro_t *m)
...
+   if (pos > m->args.count && !m->error) {
+      error_at(&state.last_loc, "macro '%pi' requires exactly %d arguments, %d were given",
+               m->name, m->args.count, pos);
       m->error = true;
    }
+   else if (pos != m->args.count) {
+      for (; pos < m->args.count; pos++) {
+         if (! tb_is_empty(AGET(m->defs, pos)))
+            hash_put(macro_args, AGET(m->args, pos), xstrdup(tb_get(AGET(m->defs, pos))));
+         else if (!m->error) {
```
Anselmo95, line 619: This check suppress errors on the next time the preprocessor tries to expand a macro. I left it there but it will hide some errors when you use a macro incorrectly multiple times

Anselmo95:
Thanks for adding the feature @nickg !

PR #1524, opened 2026-04-30
vhpi: support multi-dimensional non-homogeneous array signals    

gitmodimo:
Add VHPI support for drilling into signals whose type is a
  multi-dimensional `array of … of array of record` (or any other
  non-homogeneous element). Previously `vhpi_get_value` /                                                                                                                                                                                         
  `vhpi_put_value` on `dut.X(0)(0)(0).r` SIGSEGV'd inside                                                                                                                                                                                         
  `child_scope_at`, because the runtime represents every dimension as                                                                                                                                                                             
  a single flat `SCOPE_ARRAY` of leaf-element scopes (see                                                                                                                                                                                         
  `lower_sub_signals` — `type_elem_recur` + total flat length), but                                                                                                                                                                               
  `vhpi_get_scope_prefixedName` was walking one dimension at a time                                                                                                                                                                               
  and trying to `child_scope_at` into intermediate sub-scopes that                                                                                                                                                                                
  don't exist.                                                                                                                                                                                                                                    
                                                                                                                                                                                                                              
                  
  When resolving the scope of a `c_indexedName`, skip past any                                                                                                                                                                                    
  consecutive `c_indexedName` ancestors to the anchor (the outer
  `objDecl` or `selectedName`) and index that anchor's scope once,                                                                                                                                                                                
  using this indexedName's cumulative `offset` divided by the leaf                                                                                                                                                                                
  type's `numElems`. The same formula gives the right index for any
  depth, and reduces to `BaseIndex` in the existing 1-D case                                                                                                                                                                                      
  (`offset = BaseIndex * numElems`).                                                                                                                                                                                                              
                                                                                                                                                                                                                                                  
  `c_selectedName` handling is unchanged; the recursion split keeps                                                                                                                                                                               
  indexed/selected on independent paths.
                                                                                                                                                                                                                                                  
  ## Test         
                                                                                                                                                                                                                                                  
  `test/regress/vhpi20.vhd` declares a 2×2×2 `array of array of array
  of record` signal. `test/vhpi/vhpi20.c` round-trips both
  `vhpi_get_value` and `vhpi_put_value` on:                                                                                                                                                                                                       
   
  - `x(1)(1)(1).r` — flat scope index 7 of 8                                                                                                                                                                                                      
  - `x(0)(1)(0).r` — flat scope index 2 of 8
                                                                                                                                                                                                                                                  
  On master the first `vhpi_get_value` segfaults in `child_scope_at`;                                                                                                                                                                             
  with this change both succeed.

### DIFF
--- a/src/vhpi/vhpi-model.c
+++ b/src/vhpi/vhpi-model.c
@@ -1701,26 +1701,56 @@ static rt_scope_t *vhpi_get_scope_prefixedName(c_prefixedName *pn)
    if (pn->scope)
       return pn->scope;
 
-   rt_scope_t *parent = NULL;
-
-   c_prefixedName *ppn = is_prefixedName(pn->Prefix);
-   if (ppn != NULL)
-      parent = vhpi_get_scope_prefixedName(ppn);
-
-   c_objDecl *od = is_objDecl(pn->Prefix);
-   if (od != NULL)
-      parent = vhpi_get_scope_objDecl(od);
-
-   if (parent == NULL)
-      return NULL;
-
    c_vhpiObject *obj = &(pn->name.expr.object);
    c_indexedName *in = is_indexedName(obj);
    c_selectedName *sn = is_selectedName(obj);
-   if (in != NULL)
-      pn->scope = child_scope_at(parent, in->BaseIndex);
-   else if (sn != NULL)
+
+   if (in != NULL) {
+      // The runtime collapses a multi-dimensional non-homogeneous
+      // array (e.g. `array of array of record`) into one flat
+      // SCOPE_ARRAY whose children are the leaf element scopes —
+      // see lower_sub_signals in lower.c which uses type_elem_recur
+      // and the total flat length.  Walk past any consecutive
+      // indexedName ancestors to the anchor scope and index directly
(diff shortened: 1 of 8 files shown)


gitmodimo:
Added a second commit expanding `vhpi20` to cover all four combinations of array style × element type:                            
                                                         
  | | Pre-VHDL-2008 nested | Native VHDL-2008 2-D |                                                                                 
  |---|---|---|
  | **Record element** | `x` (existing) | `y` (new) |                                                                               
  | **Raw SLV element** | `z` (new) | `w` (new) |        

  For `y` (native 2-D of record), `vhpiIndexedNames` expands all dimension combinations into a flat list — a single                 
  `vhpi_handle_by_index` call reaches any element directly, then `vhpi_handle_by_name` picks the record field as usual.
                                                                                                                                    
  For `z`/`w` (homogeneous SLV elements), no scope lookup is needed — `vhpi_get/put_value` selects the element via the `offset`     
  field using pure arithmetic.
                                                                                                                                    
  Each case exercises both `vhpi_get_value` and `vhpi_put_value` with `vhpiDepositPropagate`.

PR #1536, opened 2026-05-12
fail gracefully in resolve_record_ref_or_call upon referencing library.

### LINKED ISSUE #1535: Exception when missing `entity` keyword in direct instantiation
I get a crash when trying to instantiate an entity, but forgot the `entity` keyword.
```vhdl
library ieee;
use ieee.std_logic_1164.all;
use ieee.numeric_std.all;

--hdlregression:tb
entity test is
end entity;

architecture rtl of test is
begin
  i_test : work.test1;
end architecture;
```
Executed in a clean folder I get
```
$ nvc --std=08 -a test_missing_entity.vhd
*** Caught exception c0000005 (EXCEPTION_ACCESS_VIOLATION) [address=0000000000000000, ip=00007FF6CD14DDD8] ***

[00007ff6ccf612b8] _nvc_file_canseek+0x12e8
[00007ff83a1458a2] (KERNELBASE) UnhandledExceptionFilter+0x1f2
[00007ff83c80a512] (ntdll) strncpy+0x2682
[00007ff83c7c1712] (ntdll) _C_specific_handler+0x92
[00007ff83c80623e] (ntdll) _chkstk+0x9e
[00007ff83c6b2396] (ntdll) RtlLocateExtendedFeature+0x596
[00007ff83c805b7d] (ntdll) KiUserExceptionDispatcher+0x2d
[00007ff6cd14ddd7] nvc_random_get_next+0x7127
[00007ff6cd14dae4] nvc_random_get_next+0x6e34
[00007ff6ccffdf52] std_env_get_assert_format+0x64502
[00007ff6ccfc5659] std_env_get_assert_format+0x2bc09
[00007ff6ccfc1bee] std_env_get_assert_format+0x2819e
[00007ff6ccfbe73c] std_env_get_assert_format+0x24cec
[00007ff6ccf6f04f] polyfill_to_ostring_bit_vec+0x7aef
[00007ff6ccf5a99f]
[00007ff6ccf54219]
[00007ff6ccf410fd]
[00007ff6ccf413e5]
[00007ff83b89e8d6] (KERNEL32) BaseThreadInitThunk+0x16
[00007ff83c72c3fb] (ntdll) RtlUserThreadStart+0x2b
```

With --enable-debug set in config I get the following
```
$ nvc --std=08 -a test_missing_entity.vhd
** Debug: C runtime library: C:\WINDOWS\System32\ucrtbase.dll
Assertion failed: item->object != NULL, file ../src/tree.c, line 682
```

```
Win 11 clang64
nvc 1.21-devel (1.20.0.r78.gda776e5a4) (Using LLVM 22.1.4)
```

Blebowski:
Resolve library reference in `resolve_record_ref_or_call` in case the record prefix being referenced is a library.

Fixes #1535.

I did not know how to fix this on the parser level since the reproducer is de-facto valid syntax for start of record field assignment, e.g.:
```
label : my_rec.my_sig <= '1';
```

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -13182,6 +13182,38 @@ static tree_t p_psl_or_concurrent_assert(ident_t label)
    return conc;
 }
 
+static bool is_component_instantiation_statement(ident_t label)
+{
+   const token_t p2 = peek_nth(2);
+
+   if (label == NULL)
+      return false;
+
+   if (p2 == tSEMI || p2 == tGENERIC || p2 == tPORT)
+      return true;
+
+   if (p2 != tDOT)
+      return false;
+
+   const token_t p4 = peek_nth(4);
+   const token_t p6 = peek_nth(6);
+   const token_t p7 = peek_nth(7);
+   bool rv = false;
+
+   if (p4 == tSEMI || p4 == tGENERIC || p4 == tPORT || p4 == tEOF)
+      rv = true;
+
+   if (p4 == tLPAREN && p6 == tRPAREN &&
+       (p7 == tSEMI || p7 == tGENERIC || p7 == tPORT || p7 == tEOF))
+      rv = true;
+
+   if (rv)
+      parse_error(CURRENT_LOC,
+                  "missing 'entity' keyword in entity binding indication.");
+
+   return rv;
+}
+
 static tree_t p_concurrent_statement(void)
 {
    // block_statement | process_statement | concurrent_procedure_call_statement
@@ -13233,8 +13265,7 @@ static tree_t p_concurrent_statement(void)
 
    case tID:
       {
-         const token_t p2 = peek_nth(2);
-         if ((label != NULL && p2 == tSEMI) || p2 == tGENERIC || p2 == tPORT)
+         if (is_component_instantiation_statement(label))
             return p_component_instantiation_statement(label, NULL);
          else {
             tree_t name = p_name(N_SUBPROGRAM), conc;
--- a/src/parse.c
+++ b/src/parse.c
@@ -10706,6 +10706,8 @@ static tree_t p_instantiated_unit(tree_t name)
          tree_set_ident2(t, tree_ident(name));
          if (tree_has_ref(name))
             tree_set_ref(t, tree_ref(name));
+         else
+            suppress_errors(nametab);
       }
       else {
          parse_error(tree_loc(name), "invalid instantiated unit name");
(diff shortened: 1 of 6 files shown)


test/parse/issue1535.vhd
```diff
@@ -0,0 +1,7 @@
+entity test is
+end entity;
+
+architecture rtl of test is
+begin
+  i_test : work.test1;
+end architecture;
```
maehne: Newline missing at end of line.

test/test_parse.c
```diff
@@ -7617,6 +7617,24 @@ START_TEST(test_issue1498)
...
 END_TEST
 
+START_TEST(test_issue1535)
+{
+   set_standard(STD_08);
+
+   input_from_file(TESTDIR "/parse/issue1535.vhd");
+
+   const error_t expect[] = {
+      { 6, "invalid procedure call" },
```
maehne: Is this error message clear enough for a user? Wouldn't be a hint helpful that the keyword `entity` might be missing? After all, `test1` is prefixed with the library `work` which hints to an entity. Procedures, components, or configurations would be typically stored in a package or declared/defined locally in the architecture. Entities and components are not required to have generics or ports.

maehne (commented):
I saw this PR by chance and the error message checked in the test1535 triggered me to leave a comment.

test/test_parse.c
Blebowski: At parser level its messy to distinguish between concurrent procedure call and instance with forgotten
entity keyword. One would need to parse the name first, look if it points to existing entity or procedure
in package, and proceed based on that. If the instantiated entity does not exist, then that would not work.

I tried to do bit of guessing, and extended the parser. It relies on the fact that the concurrent procedure
call will have one more hierarchy (the package), while signal assignment statement will have `<=` at some
point.

Not sure if this is not too messy.

Blebowski: Still the check in the names resolution needs to be there since some crazy cases like:

```
i_test : work.sub(
```
with end of file following were crashing.

maehne: Thanks for addressing my comment! The new error message is much more helpful!

src/parse.c
```diff
@@ -13182,6 +13182,38 @@ static tree_t p_psl_or_concurrent_assert(ident_t label)
    return conc;
 }
 
+static bool is_component_instantiation_statement(ident_t label)
```
nickg: This doesn't work for something like:

```vhdl
package pack is
    procedure proc;
end package;

entity test is
end entity;

use work.pack;

architecture test of test is
begin
    x: pack.proc;
end architecture;
```

nickg (approved):

PR #1555, opened 2026-05-27
Fix VHPI fields of array record elements

gitmodimo:
Fix a VHPI crash/incorrect lookup when external tools inspect a field inside a record that is stored in an array.

  For example, this kind of VHDL object could fail through VHPI:

  ```vhdl
  type sample_record_t is record
     data  : std_logic_vector;
     valid : std_logic_vector(0 downto 0);
  end record;

  type sample_array_t is array (natural range <>) of sample_record_t;

  signal sample : sample_array_t(0 to 1)(
     data(7 downto 0));
```
  A VHPI client inspecting this path:
```
  sample(0).data
```
  could get a bad size/constraint, read from the wrong memory location, or crash while NVC was building the VHPI type information.

  This change fixes that by making NVC:

  - keep the constrained field type available while building VHPI subtype metadata
  - calculate field addresses from the selected record element, not from the whole array object
  - unwrap array storage when the runtime representation uses an unconstrained array wrapper

  The existing issue1161 test also had a stale expectation. `AIn(0).Data` is 8 lanes × 16 bits = 128 bits, not 256 bits, so that check is corrected.

### DIFF
--- a/src/vhpi/vhpi-model.c
+++ b/src/vhpi/vhpi-model.c
@@ -1940,6 +1940,25 @@ static void *vhpi_get_value_ptr(c_vhpiObject *obj)
       if (base == NULL)
          return NULL;
 
+      c_objDecl *pod = is_objDecl(pobj);
+      type_t ptype = pod != NULL ? pod->decl.type : NULL;
+      if (ptype != NULL && type_is_array(ptype)) {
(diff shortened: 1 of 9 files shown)


nickg:
I think this was fixed by 4f2d0d27da262. I pushed the test.

Issue #1570, opened 2026-06-09
inout parameters in foreign VHPI function/procedure not properly handled


dalex78:
This ticket is openned following the investigation from the discussion https://github.com/nickg/nvc/discussions/1568.

There seem to be an issue with `inout` parameters in foreign VHPI function/procedure and the NVC implementation.
The VHPI function 'vhpi_get_value' seems to not take this case into account correctly. When the actual value for the parameter is called, there is no distinction made between:

- `vhpiConstParamDeclK` which store their value directly into a `jit_scalar_t` type (variant `int64_t integer` or `double real`).
- `vhpiVarParamDeclK` which store a pointer to their value into a `jit_scalar_t` type (variant `void *pointer`).

Therefore, when dealing with `inout` parameters, the current NVC implementation returns a truncated value of the pointer.

```diff
diff --git a/src/vhpi/vhpi-model.c b/src/vhpi/vhpi-model.c
index 651dab26b..378ffaa85 100644
--- a/src/vhpi/vhpi-model.c
+++ b/src/vhpi/vhpi-model.c
@@ -3454,38 +3454,44 @@ int vhpi_get_value(vhpiHandleT expr, vhpiValueT *value_p)
 
    case vhpiConstParamDeclK:
    case vhpiVarParamDeclK:
       {
          vhpi_context_t *c = vhpi_context();
          assert(c->args != NULL);
 
          c_interfaceDecl *id = is_interfaceDecl(obj);
          assert(id != NULL);
 
          if (td->wrapped) {
             value = c->args[id->argslot].pointer;
             num_elems = ffi_array_length(c->args[id->argslot + 2].integer);
          }
          else if (td->IsComposite) {
             value = c->args[id->argslot].pointer;
             num_elems = td->numElems;
          }
-         else
-            value = (void *)&(c->args[id->argslot]);
+         else {
+            if (vhpi_get_prefix_kind(obj) == vhpiConstParamDeclK) {
+               value = (void *)&(c->args[id->argslot]);
+            } else {
+               // assert(false);
+               value = (void *)(c->args[id->argslot].pointer);
+            }
+         }
       }
       break;
 
    default:
       vhpi_error(vhpiError, &(obj->loc), "class kind %s cannot be used with "
                  "vhpi_get_value", vhpi_class_str(obj->kind));
       return -1;
    }
 
    assert(td->IsComposite || num_elems == 1);
    assert(num_elems >= 0);
 
[...]

PR #1571, opened 2026-06-09
[VHPI] Fix for inout parameter in foreign VHPI function/procedure

dalex78:
Issue: https://github.com/nickg/nvc/issues/1570
Discussion: https://github.com/nickg/nvc/discussions/1568

### DIFF
--- a/src/vhpi/vhpi-model.c
+++ b/src/vhpi/vhpi-model.c
@@ -3469,8 +3469,13 @@ int vhpi_get_value(vhpiHandleT expr, vhpiValueT *value_p)
             value = c->args[id->argslot].pointer;
             num_elems = td->numElems;
          }
-         else
-            value = (void *)&(c->args[id->argslot]);
+         else {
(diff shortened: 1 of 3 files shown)


src/vhpi/vhpi-model.c
```diff
@@ -3469,8 +3469,13 @@ int vhpi_get_value(vhpiHandleT expr, vhpiValueT *value_p)
             value = c->args[id->argslot].pointer;
             num_elems = td->numElems;
          }
-         else
-            value = (void *)&(c->args[id->argslot]);
+         else {
+            if (vhpi_get_prefix_kind(obj) == vhpiConstParamDeclK) {
```
nickg, line 3473: This doesn't work as a parameter have class `variable` and mode `in`. I fixed it in 8c505aa.

dalex78: Ok, yes indeed. Thank you. for the fix. I have confirmed that my fix didn't work as well by testing that case.
Here is a PR to add new tests to catch that case: https://github.com/nickg/nvc/pull/1574

PR #1634, opened 2026-08-08
Collect code coverage for package declarations

Blebowski:
This is rebased PR for code coverage in packages from:
[https://github.com/nickg/nvc/pull/1456](https://github.com/nickg/nvc/pull/1456)

I have tried this also on my CTU CAN FD project that has many own packages and uses VUNIT, that works fine.

The above mentioned PR however mentions:
[https://github.com/MymyMouss/nvc/commit/e8bb89950d1550c566055b802f99bcf579b6c4a8](https://github.com/MymyMouss/nvc/commit/e8bb89950d1550c566055b802f99bcf579b6c4a8)

I read the AI summary, and it seemed to make some sense, though I did not understand
it fully (damn, the AI really takes away part of the understanding !).
 
The commit above does not provide reliable reproducer though, and I was not able to find the repo
referenced from that commit. Since I was lost, I fired up Claude too, and let it find a minimal reproducer.
See the other bug report.

I am not 100 % sure this is the case that was problem in the commit summary above, however, it
claims that this is not dependant on coverage in packages though. 

Possibly the person posting the AI summary, enabled coverage on packages, and then tried to
clean the coverage and exclude some unreachable code by pragmas, therefore hitting this case. 

Alternatively, his / her code had some pragmas like this pre-existing in the packages since
commercial tools have the same pragmas. Then it popped up again only with the PR that added
NVC coverage in packages.

FWIW, it might be good to enable full coverage in the External projects, Vunit and OSVVM regressions.
The run-time will be slightly higher, but may prevent regressions in the coverage engine now that it is
used by some people.

Anyway, this PR should be working coverage for VHDL packages.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,9 @@
 - Added support for System Verilog procedural assertions.
 - Fixed an issue where VHPI could not iterate the constraints of objects
   declared with an unconstrained subtype (#1632).
+- Code coverage can now be collected for package declarations, opted
+  in via the new `+package` directive in a coverage specification
+  file (#1007).
 
 ## Version 1.22.1 - 2026-08-01
 - Waveform dumping now dumps multiple top units instead of only the first
--- a/nvc.1
+++ b/nvc.1
@@ -1034,12 +1034,30 @@ NVC can collect code coverage only on part of the simulated design.
 When coverage specification file is passed during elaboration time,
 NVC collects code coverage only as specified in this file. If
 the file is ommited, NVC collects code coverage on whole design.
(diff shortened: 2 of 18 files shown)


Blebowski:
Hi @nickg,

I see that you majorly refactored the code coverage API. Do you think it is worth to try rebasing this MR ?
I so, any comments to this are appreciated.

nickg:
> 
> I see that you majorly refactored the code coverage API. Do you think it is worth to try rebasing this MR ? I so, any comments to this are appreciated.

Let's just park it for now. I want to make some other changes first.

